PR #1, opened 2025-04-01
Initial Infra for aws and rke2 clusters

igomez06:
Implemented an aws terraform module for cluster nodes, and a rke2 playbook for configuring an rke2 cluster. There are README files documenting how to run each piece.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+.DS_Store
--- a/README.md
+++ b/README.md
@@ -1,1 +1,12 @@
-# qa-infra-automation
+# Infrastructure Automation Repository
+
+This repository contains infrastructure automation scripts using Terraform and Ansible. It is organized into directories for each tool and specific infrastructure components.
+
+## Current Directory Structure
+
+├── ansible/
+│   └── rke2/
+│       
+├── terraform/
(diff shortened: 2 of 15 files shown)


git-ival (changes requested):
We should avoid including ARNs for internal resources wherever possible. We can just use stubs instead, like vpc_id=<desired vpc id> or something

igomez06:
> We should avoid including ARNs for internal resources wherever possible. We can just use stubs instead, like vpc_id= or something

fixed

git-ival:
I couldn't tell after reviewing the ansible playbook, is there any output that generates/pulls the clusters kubeconfig locally? I think it would be useful if this was a feature, so this can be more easily used outside of Jenkins if desired.

igomez06:
> I couldn't tell after reviewing the ansible playbook, is there any output that generates/pulls the clusters kubeconfig locally? I think it would be useful if this was a feature, so this can be more easily used outside of Jenkins if desired.

Yes there is in rke2 playbook check lines 50-58

git-ival (approved):
LGTM!

caliskanugur (approved):
LGTM

PR #2, opened 2025-04-07
Ansible Rancher Playbook

igomez06:
Updated the gitignore. Found some issues with the rke2 scripts with different role configuration, so fixed those. Also created the rancher-playbook.yaml for install rancher on clusters.

### DIFF
--- a/ansible/rancher/rancher-playbook.yml
+++ b/ansible/rancher/rancher-playbook.yml
@@ -1,0 +1,101 @@
+---
+- hosts: localhost
+  connection: local
+  gather_facts: false
+  vars_files:
+    - vars.yaml # Assuming this file contains rancher_version and cert_manager_version
+  vars:
+    # Get the current Terraform workspace name
+    terraform_workspace: "{{ lookup('env', 'TF_WORKSPACE') }}"
+    # Construct the path to the workspace's state file
+    terraform_state_file: "../../terraform/aws/cluster_nodes/terraform.tfstate.d/{{ terraform_workspace }}/terraform.tfstate"
+    fqdn: "{{ lookup('cloud.terraform.tf_output', 'fqdn', state_file=terraform_state_file) }}"
+  tasks:
+    - name: Check if kubeconfig file exists
+      stat:
+        path: "{{ kubeconfig_file }}"
+      register: kubeconfig_file_stat
+
+    - name: Fail if kubeconfig file does not exist
+      fail:
+        msg: "kubeconfig file '{{ kubeconfig_file }}' does not exist."
+      when: not kubeconfig_file_stat.stat.exists
+
+    - name: Install cert-manager CRDs
+      kubernetes.core.k8s:
+        kubeconfig: "{{ kubeconfig_file }}"
+        src: https://github.com/cert-manager/cert-manager/releases/download/v{{ cert_manager_version }}/cert-manager.crds.yaml
+        state: present
+      when: cert_manager_version != ""
+
+    - name: Install cert-manager
+      kubernetes.core.k8s:
+        kubeconfig: "{{ kubeconfig_file }}"
+        src: https://github.com/cert-manager/cert-manager/releases/download/v{{ cert_manager_version }}/cert-manager.yaml
+        state: present
+        wait: true
+        wait_timeout: 300
+      when: cert_manager_version != ""
+
+    - name: Wait for cert-manager to be ready
+      kubernetes.core.k8s_info:
+        kubeconfig: "{{ kubeconfig_file }}"
+        api_version: apps/v1
+        kind: Deployment
+        namespace: cert-manager
+        name: cert-manager
+      register: cert_manager_deployment
+      until: cert_manager_deployment.resources[0].status.readyReplicas == cert_manager_deployment.resources[0].status.replicas
+      retries: 10
+      delay: 30
+      when: cert_manager_version != ""
+
+    - name: Add Rancher Helm repository
+      kubernetes.core.helm_repository:
+        kubeconfig: "{{ kubeconfig_file }}"
+        name: "{{ rancher_chart_repo }}"
+        repo_url: "{{ rancher_chart_repo_url }}"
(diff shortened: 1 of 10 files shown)


ansible/rancher/rancher-playbook.yml
```diff
@@ -0,0 +1,101 @@
...
+
+    - name: Wait for Rancher to be ready
+      kubernetes.core.k8s_info:
+        kubeconfig: "{{ kubeconfig_file }}"
+        api_version: apps/v1
+        kind: Deployment
+        namespace: cattle-system
+        name: rancher
+      register: rancher_deployment
+      until: rancher_deployment.resources[0].status.readyReplicas == rancher_deployment.resources[0].status.replicas
```
caliskanugur, line 83: Is there a way to get rid of indexing for the resources? Off the top of my head, I want to say `kubectl` and `jq` or `yq`, but these are host-dependent, and hosts may not have them?

ansible/rancher/rancher-playbook.yml
```diff
@@ -0,0 +1,101 @@
...
+        chart_ref: "{{ rancher_chart_repo }}/rancher"
+        release_namespace: cattle-system
+        create_namespace: true
+        chart_version: "{{ rancher_version }}"
+        wait: true
+        wait_timeout: "600s"
+        values:
+          hostname: "{{ fqdn }}"
+          bootstrapPassword: "{{ bootstrap_password }}"
+      when: rancher_version != ""
```
caliskanugur, line 73: Seems like this `rancher_version` is used both for helm repo configuration and helm installation, same for `cert_manager_version` or other conditions:

If we had a host that already had this repo configuration set, would it still run that step?

Is there a way to enhance these `when` conditions, something for the future - especially if we decide to have more granular playbooks, f.e, airgap and rancher installation playbooks that are calling a helm-configure-and-install playbook

caliskanugur (approved):
LGTM!

I've got a few questions that are not in the scope of this PR

ansible/rancher/rancher-playbook.yml
```diff
@@ -0,0 +1,101 @@
...
+      kubernetes.core.helm:
+        kubeconfig: "{{ kubeconfig_file }}"
+        name: rancher
+        chart_ref: "{{ rancher_chart_repo }}/rancher"
+        release_namespace: cattle-system
+        create_namespace: true
+        chart_version: "{{ rancher_version }}"
+        wait: true
+        wait_timeout: "600s"
+        values:
```
git-ival, line 70: Maybe not for this PR, but the ability to set arbitrary values for the Rancher helm chart may be necessary. Especially when it comes to `extrEnv` vars. For example, the latest experimental cluster role aggregation feature that went out in 2.11.0 can -only- be enabled at install-time, which requires setting the correct `extraEnv` var

ansible/rancher/rancher-playbook.yml
git-ival, line 83: Should we add a check to see if the healthcheck endpoints of the ingress is in a good state? https://ranchermanager.docs.rancher.com/how-to-guides/new-user-guides/infrastructure-setup/amazon-elb-load-balancer#health-check-paths-for-nginx-ingress-and-traefik-ingresses

git-ival (approved):
LGTM overall, just a few suggestions/questions but they can be handled in a follow-up PR if they need to be done 🚀

ansible/rancher/rancher-playbook.yml
igomez06: No especially if we 're using the The _**kubernetes.core.k8s_info**_ module, this module returns a dictionary with a resources list to provide a consistent way to handle queries that might return one object or many. So in essence it's a safe call, if we can't index then the deployment failed anyway right? This a much cleaner way than installing jq, running a raw bash command to run kubectl and then parsing the output.

ansible/rancher/rancher-playbook.yml
igomez06: We can definitely do that, I think for now we should be good, but yes I agree.

ansible/rancher/rancher-playbook.yml
igomez06: Even if we had a repo set we need the version to what we're installing. Even for air gap we take a version for either. I guess I'm not understanding what the concern is.

ansible/rancher/rancher-playbook.yml
git-ival: Should be as simple as `curl {{fqdn}/ping` or `curl {{fqdn}}/healthz` and making sure the response status is 200 I guess. Not sure if that will work that easily but if its handled elsewhere then its probably not needed

git-ival: Actually I guess it would be after the "Display Rancher URL" step with `curl https://{{ rancher_ingress.resources[0].spec.rules[0].host }}/[ping || healthz]`

igomez06: aaah yes I can do that. Let me try, before I merge the PR.

ansible/rancher/rancher-playbook.yml
caliskanugur: So what I am asking is, why does it have to use the same condition both for configuring and installing the helm chart? These two are separate tasks, but their condition is the same

- If the variable is not empty and the helm repo is not configured -> configure the helm repo
- If the variable is not empty and helm repo is not already installed on the cluster -> install the helm chart

Are these even possible to configure?

ansible/rancher/rancher-playbook.yml
caliskanugur: From the upstream documentation:

> Note:
> Ingress is frozen. New features are being added to the [Gateway API](https://kubernetes.io/docs/concepts/services-networking/gateway/).

https://kubernetes.io/docs/concepts/services-networking/ingress 

So, even if we added this condition, we may have to update it soon. In my opinion it's not really worth to add since it's lifecycle gonna be relatively short

PR #5, opened 2025-05-20
Fixes for bugs found when running in Jenkins

igomez06:
Updated pathing, as well as added new more efficient checks in both the rke2 and rancher playbooks. Also updated READMEs.


ansible/rke2/rke2-playbook.yml
```diff
@@ -108,4 +110,27 @@
...
+    - vars.yaml
+  tasks:
+    - name: Check if all RKE2 pods are running
+      kubernetes.core.k8s_info:
+        kubeconfig: "{{ kubeconfig_file }}"
+        api_version: v1
+        kind: Pod
+        namespace: kube-system
+        label_selectors:
+          - app.kubernetes.io/managed-by=Helm
```
rancher-max, line 128: I don't remember for sure, but I believe there are some system-level pods that don't have this label. That being said, if they are failing, then generally there is at least one pod with this label that will also be failing, so this should be sufficient. Just wanted to note in case we see some weirdness ever.

ansible/rancher/README.md
```diff
@@ -0,0 +1,52 @@
+# Running the Rancher Playbook
```
rancher-max, line 1: It might be helpful to include an example in this at some point (with dummy values instead of any sensitive info, if there is any required). This is great for right now though!

rancher-max (approved):
Left some general comments but overall LGTM!

ansible/rke2/README.md
```diff
@@ -14,14 +14,16 @@ This playbook deploys an RKE2 Kubernetes cluster.
 ## Installation
 
 1.  **Install Ansible and its dependencies:**
-
     ```bash
-    pip install --include-deps ansible
-    pip install ansible-inventory-terraform
+    ansible-galaxy collection install cloud.terraform
```
caliskanugur, line 18: Question here,

Do you think the same applies here to run this standalone?

```
*   **Ansible:** Version 2.9 or later is recommended.
*   **Python 3.11:**  Along with the `ansible` and `kubernetes` packages.
```

From the Rancher playbook?

caliskanugur (commented):
LGTM, a few questions

ansible/rke2/rke2-playbook.yml
rancher-max: Yes that should work! Good call. Created https://github.com/rancher/qa-tasks/issues/1842 and just put it in both PIT and RAFT projects for one of us to take.

caliskanugur (approved):

rancher-max (approved):

PR #8, opened 2025-06-20
tofu by default. add downstream infra

slickwarren:
* switched from terraform -> tofu
* added infra option for downstream (from rancher)


tofu/rancher/cluster/main.tf
```diff
@@ -0,0 +1,177 @@
...
+}
+
+resource "random_string" "suffix" {
+  length  = 3
+  upper   = false
+  special = false
+}
+
+
+resource "rancher2_cloud_credential" "rancher2_cloud_credential" {
```
git-ival, line 14:
```suggestion
module "cloudcredential" {
  source = "../cloudcredential"
  ...
}
```
Can we use a "module" block instead of copy/pasting the `cloudcredential` module here?

tofu/rancher/cluster/main.tf
```diff
@@ -0,0 +1,177 @@
...
+    for_each = var.cloud_provider == "harvester" ? [1] : []
+    content {
+      cluster_id         = var.node_config.cluster_v1_id
+      cluster_type       = var.node_config.cluster_type
+      kubeconfig_content = var.node_config.kubeconfig_content
+    }
+  }
+}
+
+resource "rancher2_machine_config_v2" "rancher2_machine_config_v2" {
```
git-ival, line 42: Can we move this into its own module and do the same as my comment above? It would be nice to be able to just pass inputs to a component module block instead of maintaining separate instances of this "dynamic" resource for different modules.

git-ival:
I made a couple of suggestions, but I think generally we can basically adopt the concepts I used here https://github.com/git-ival/rancher2-scaling/tree/WIP/rancher-cluster-operations. 

Where different resources that have lots of config options have their own modules with "dynamic" blocks and the input vars dictate what gets created/configured. Then we can just call those modules via "module" blocks to piece together whatever we need. Maybe others don't agree with this approach though, so I'm fine with keeping this PR as-is if there are differing opinions on that 🫡

git-ival (approved):
Thanks for taking my suggestion! These changes should make things a lot more maintainable and reusable going forward 🚀

PR #9, opened 2025-06-21
tofu by default

slickwarren:


tofu/aws/cluster_nodes/terraform.tf
```diff
@@ -0,0 +1,13 @@
+terraform {
+  required_version = ">= 0.13.1"
```
git-ival, line 2: Is this the correct opentofu version?

igomez06: Is this the version we should be using? 1.8.2

slickwarren: updated the minimum version based on offline discussion

igomez06 (approved):
LGTM

git-ival (approved):
🚀

PR #13, opened 2025-07-17
Custom Cluster Provisioning

slickwarren:
also enhances existing packages:
* harvester tf module

introduces:
* terraform package that provisions (empty) downstream custom cluster
* ansible package that registers nodes to said cluster using the output from the aforementioned tf package


tofu/rancher/custom_cluster/README.md
```diff
@@ -0,0 +1,41 @@
...
+## Prerequisites
+
+* An api_key from your rancher setup
+* tofu installed on your client machine
+
+## Usage
+
+
+1.  **Create the downstream cluster**
+    * see the [variables section](#sample) to configure the cluster
```
rancher-max, line 14: This doesn't seem to point anywhere

tofu/rancher/custom_cluster/README.md
```diff
@@ -0,0 +1,41 @@
...
+2.  **Destroy the downstream cluster:**
+
+    ```bash
+    tofu -chdir=tofu/rancher/custom_cluster destroy -auto-approve -var-file=/path/to/vars.tfvars
+    ```
+
+    Use the same `vars.tfvars` file or `-var` flags used during `apply`.
+
+## Outputs
+Refer to [outputs.tf](./outputs.tf) for a list of exported values.
```
rancher-max, line 30: nit: Since there's only one exported value, it might be nice to put it here into more "human-readable" terms

tofu/rancher/custom_cluster/variables.tf
```diff
@@ -0,0 +1,36 @@
+variable "kubernetes_version" {}
```
rancher-max, line 1: It would be helpful if all of the variables here had descriptions so it is clear when and why they should be used

tofu/rancher/custom_cluster/outputs.tf
```diff
@@ -0,0 +1,4 @@
+output "cluster_registration_token" {
+  value = data.rancher2_cluster_v2.rancher2_cluster_v2.cluster_registration_token[0]["insecure_node_command"]
```
rancher-max, line 2: We only want the insecure command to be output if that variable is set to true, right? Otherwise it should be the command without the insecure flag

tofu/harvester/modules/vm/variables.tf
```diff
@@ -54,6 +70,8 @@ variable "namespace" {
   default = "default"
 }
 
-variable "hostname_prefix" {
-  description = "prefix to append to created resources"
+variable "generate_name" {
```
rancher-max, line 73: Can we keep the description in here?

tofu/harvester/modules/vm/main.tf
```diff
@@ -4,7 +4,7 @@ locals {
   temp_node_names = flatten([
     for node_group in var.nodes : [
       for i in range(node_group.count) : {
-        name =  "${var.hostname_prefix}-${join("-", node_group.role)}-${i}"
+        name =  "${join("-", node_group.role)}-${i}"
```
rancher-max, line 7: Will this not have the name specified in variables prefixed to it anymore?

tofu/harvester/modules/vm/main.tf
```diff
@@ -34,53 +34,39 @@ resource "random_string" "random_suffix" {
...
-    "userdata" = local.cloud_init
+    "userdata" = "${var.cloud_init}\nssh_authorized_keys:\n  - ${var.ssh_key}\n"
   } 
 }
 
-
 resource "harvester_virtualmachine" "vm" {
-
   for_each = { for node in local.node_names : node.name => node }
+  # increate timeout to allow for updates
```
rancher-max, line 58: nit: typo in the comment

tofu/harvester/modules/vm/main.tf
```diff
@@ -119,7 +105,20 @@ resource "harvester_virtualmachine" "vm" {
...
+  name = each.value.name
+  variables = {
+    # Connection vars.
+    ansible_user = var.ssh_user
+    ansible_host = harvester_virtualmachine.vm[each.key].network_interface[0].ip_address
+    ansible_role = join(",", each.value.role)
+
+  }
+  depends_on = [harvester_virtualmachine.vm]
+}
```
rancher-max, line 124: If ansible is not run directly after this, does having this present any issues? Also, could I run the ansible if I didn't start with this tofu module? It might be helpful to add a comment here explaining.

ansible/rancher/default-ha/cleanup.sh
rancher-max, line 1: Is this file necessary? Or can the different cleanup scripts just be run directly from their own directories? It's throwing me off a little because running `.../default-ha/cleanup.sh` does not do any cleanup to the default-ha setup, versus `.../downstream/custom_cluster/cleanup.sh` does cleanup that setup.

ansible/rancher/downstream/custom_cluster/cleanup.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+set -e
+
+. vars.env
+
+# destroy empty cluster
+tofu -chdir="$TERRAFORM_CLUSTER_SOURCE" destroy -auto-approve -var-file="$TFVARS_FILE" &
+
+# destroy nodes
+tofu -chdir="$TERRAFORM_NODE_SOURCE" destroy -auto-approve -var-file="$TFVARS_FILE" &
```
rancher-max, line 10: I'm not sure this file is necessary within ansible. It seems to just be a script that is coupled to tofu -- but what if I provided VMs separately and didn't manage them via tofu? Perhaps we can just include this in the relevant Tofu module(s)' README(s).

ansible/rke2/default/cleanup.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+set -e
+
+. vars.env
+
+# destroy nodes
+tofu -chdir="$TERRAFORM_NODE_SOURCE" destroy -auto-approve -var-file="$TFVARS_FILE"
```
rancher-max, line 7: Same comment as on the other cleanup script

rancher-max (changes requested):
The ansible here seems to be highly coupled with the tofu. Is it possible to have it decoupled so that I can create some VMs manually and then run this ansible to have them become custom clusters in Rancher?

ansible/rancher/default-ha/cleanup.sh
slickwarren: yeah I can remove it. It was just something I was playing around with tbh.

tofu/harvester/modules/vm/main.tf
slickwarren: 1. no, you can use the resources created from main.tf in whatever capacity you want after this. My understanding is, this is just setting up something that ansible can access, but doesn't break anything if you don't use ansible / these vars.
2. right now, no. This would need to be decoupled in some way in order to get ansible to run independently of terraform. 

again, this is modeling behavior from tofu/aws/modules/cluster_nodes

tofu/rancher/custom_cluster/outputs.tf
slickwarren: hmm. Can we leave this for improvement later? Right now, the ansible scripts are setup to explicitly use cert-manager self signed certs (insecure). So while I understand the ask to not hard code this to the insecure command, for now I feel it makes the most sense to leave it as insecure. Also having it not hard coded will make getting the correct output a little more challenging.

tofu/rancher/custom_cluster/README.md
rancher-max: I think it's because the heading below's name is `` Sample `vars.tfvars` `` so just the sample part doesn't point directly to it

tofu/rancher/custom_cluster/outputs.tf
rancher-max: Sure, I created a task so that it doesn't get lost: https://github.com/rancher/qa-tasks/issues/1909

tofu/harvester/modules/vm/main.tf
rancher-max: Sure I'm fine with this being in the overarching "improve the things" task. For now the consistency is good.

tofu/harvester/modules/vm/main.tf
rancher-max: We talked about the coupling of ansible and tofu offline and decided to decouple as a separate effort. Thank you!

rancher-max (approved):
We talked offline about decoupling the tf from the ansible and decided it can be a separate work effort. This LGTM -- the one comment on the README not pointing is a nit so doesn't need to be fixed directly here if you don't want to.

tofu/rancher/custom_cluster/README.md
slickwarren: I meant to fix this before 😂

lscalabrini01 (approved):

PR #14, opened 2025-08-05
Add tarball install for airgapped K8S cluster

floatingman:
This creates a set of roles and playbooks to install RKE2 on an airgap cluster through a bastion server.


ansible/rke2/airgap/docs/CNI_CONFIGURATION_GUIDE.md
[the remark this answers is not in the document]
floatingman: Sounds good

ansible/rke2/airgap/docs/CNI_CONFIGURATION_GUIDE.md
```diff
@@ -0,0 +1,373 @@
...
+- **Kubernetes Network Policies**: https://kubernetes.io/docs/concepts/services-networking/network-policies/
+
+## Version Compatibility
+
+| RKE2 Version | Canal | Calico | Cilium | Multus |
+|--------------|-------|--------|--------|--------|
+| v1.31.x      | ✅    | ✅     | ✅     | ✅     |
+| v1.30.x      | ✅    | ✅     | ✅     | ✅     |
+| v1.29.x      | ✅    | ✅     | ✅     | ✅     |
+
```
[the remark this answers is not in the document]
floatingman: Sure

ansible/rke2/airgap/playbooks/diagnose-registry.yml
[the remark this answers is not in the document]
floatingman: Sounds good

ansible/rke2/airgap/README.md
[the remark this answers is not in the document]
floatingman: This is covered in the Ansible documentation.

ansible/rke2/airgap/docs/CNI_CONFIGURATION_GUIDE.md
```diff
@@ -0,0 +1,373 @@
...
+    # Enable Calico network policies
+    network_policy: true
+```
+
+**Features:**
+- ✅ Network policies (via Calico)
+- ✅ Simple configuration
+- ✅ Good performance
+- ✅ Wide compatibility
+- ❌ Limited advanced features
```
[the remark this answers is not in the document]
floatingman: If it's not too distracting to others, I'd like to keep this.

ansible/rke2/airgap/README.md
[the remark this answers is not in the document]
slickwarren: right, but still I think many of us would be new to this. So I think it would be nice to have a link to their docs. Thoughts?

ansible/rke2/airgap/inventory/group_vars/all.yml.template
```diff
@@ -0,0 +1,92 @@
...
+    - rke2-snapshot-controller
+    - rke2-snapshot-controller-crd
+    - rke2-snapshot-validation-webhook
+rke2_agent_options: ""
+
+# Installation Method Configuration
+# Available methods:
+#   - "tarball": Uses pre-downloaded RKE2 tarballs with embedded images (no registry required)
+#     Playbook: rke2-tarball-playbook.yml
+#     Best for: Pure airgap environments, simple deployments
```
rancher-max, line 20: I don't think we need to include these comments since they are mentioned multiple times in the README as well

ansible/rke2/airgap/inventory/group_vars/all.yml.template
```diff
@@ -0,0 +1,92 @@
...
+cluster_dns: "10.43.0.10"
+
+# CNI Configuration
+# Available CNI options:
+#   - "canal" (default): Flannel + Calico for network policy (RKE2 default)
+#   - "calico": Calico CNI for advanced networking and network policies
+#   - "cilium": Cilium CNI for advanced networking, security, and observability
+#   - "multus": Enable Multus for multiple network interfaces (can be combined with others)
+#   - "none": Disable CNI installation (bring your own CNI)
+cni_plugin: "canal"  # Default RKE2 CNI
```
rancher-max, line 39: can we change this variable to just be `cni` to match rke2? And then also I think let's remove the comment since it is in other docs here already and rke2 docs should actually be the source of truth on that (for example, `multus,calico` should be a valid option, but not `multus` on its own)

ansible/rke2/airgap/inventory/group_vars/all.yml.template
```diff
@@ -0,0 +1,92 @@
...
+  canal:
+    # Flannel backend: vxlan, host-gw, udp
+    flannel_backend: "vxlan"
+    # Enable Calico network policies
+    network_policy: true
+    
+  # Multus specific settings
+  multus:
+    # Default CNI to use with Multus (canal, calico, cilium)
+    default_cni: "canal"
```
rancher-max, line 73: Is this required? I feel like the other option should just pass into the rke2 config.yaml directly

ansible/rke2/airgap/playbooks/deploy/rke2-tarball-playbook.yml
```diff
@@ -0,0 +1,214 @@
...
+         {% endif %}
+
+   - name: Display final instructions
+     debug:
+       msg: |
+         🎉 RKE2 Cluster Setup Complete!
+         
+         ✅ RKE2 cluster deployed on airgap nodes
+         ✅ kubectl installed on bastion node
+         ✅ KUBECONFIG copied and configured
```
rancher-max, line 198: How do the emojis look in a shell window? It looks good here in review, but I'm curious how it will look from command prompt if I run there, and how it will look in Jenkins' HTML console logs?

ansible/rke2/airgap/playbooks/setup/setup-agent-nodes.yml
```diff
@@ -0,0 +1,73 @@
+---
+- name: Setup RKE2 Agent Nodes
```
rancher-max, line 2: Is this playbook's purpose just to add agent nodes to an existing setup? If that's the case, I wonder if this should just be a role instead.

ansible/rke2/airgap/roles/rke2_tarball/tasks/verify_checksums.yml
```diff
@@ -0,0 +1,133 @@
+---
```
rancher-max, line 1: I don't know that we'll ever really need to verify checksums. It's not incorrect or bad to have this, but just a note that it might be superfluous.

ansible/rke2/airgap/roles/rke2_upgrade/tasks/main.yml
```diff
@@ -0,0 +1,409 @@
...
+      debug:
+        msg: |
+          Current RKE2 Version: {{ current_rke2_version.stdout }}
+          Target RKE2 Version: {{ rke2_version }}
+          Upgrade needed: {{ current_rke2_version.stdout != rke2_version }}
+
+    - name: Skip upgrade if versions match
+      debug:
+        msg: "RKE2 is already at version {{ rke2_version }}. Skipping upgrade."
+      when: current_rke2_version.stdout == rke2_version
```
rancher-max, line 25: This could be confusing because we run the rke2-upgrade task, expecting it to do something, and instead it just skips. Maybe we need it to error out in the case where the version matches already?

ansible/rke2/airgap/roles/ssh_setup/tasks/main.yml
```diff
@@ -0,0 +1,141 @@
...
+    mode: '0700'
+    owner: "{{ ansible_user }}"
+    group: "{{ ansible_user }}"
+  become: true
+
+- name: Copy SSH private key with correct filename
+  copy:
+    src: "{{ ssh_private_key_file }}"
+    dest: "~{{ ansible_user }}/.ssh/{{ ssh_private_key_file | basename }}"
+    mode: '0600'
```
rancher-max, line 15: Private key I believe should be `0400` -- I think there is a restriction in AWS if using more permissive than this.

rancher-max (commented):
Couple of points:

- There are a lot of docs, which isn’t a bad thing, but it might be overkill. I think all we should have is how to run it and everything else can be included in the docs folder like you have but isn’t strictly required and may not be read.
- Do we need to have config.yaml templates, or can these be specified directly within one variable or something? Basically, what do I do if I want to pass other variables that are not defined within the template? Or similarly, what if I don't want to pass all the configuration options that are defined here in these templates?
- We talked about this offline, but looks like this currently supports a single server and potentially multiple agents. It would be nice to have a task to enhance this to support multiple servers as well.
- Also will need a task for adding rancher airgap

Overall this is really good -- I love the use of ansible roles here and the level of detail. It seems like everything is very buttoned up, just a couple of points above and some comments directly on the files, but the direction here is great! I think we should also probably get some buy-in from distros team as this currently is targeted at rke2. It might make sense to add their team to CODEOWNERS for the ansible/rke2 directory.

ansible/rke2/airgap/README.md
```diff
@@ -0,0 +1,470 @@
...
+
+### Multi-Node Clusters
+
+The system supports multi-node clusters:
+- First node in `airgap_nodes` becomes the server (control plane)
+- Additional nodes become agents
+- Token sharing is handled automatically
+- SSH proxy configuration ensures proper communication
+
+## Documentation
```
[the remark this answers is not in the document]
floatingman: Gotcha

ansible/rke2/airgap/README.md
```diff
@@ -0,0 +1,470 @@
...
+- **Playbook**: `playbooks/deploy/rke2-dockerhub-playbook.yml`
+- **Best for**: Testing, development environments with internet access
+- **Features**: Pulls images directly from Docker Hub (requires internet)
+- **Status**:  **In Progress**
+
+### 4. **Dynamic Images Method**
+- **Playbook**: `playbooks/deploy/rke2-dynamic-images-playbook.yml`
+- **Best for**: Latest versions, automated image discovery
+- **Features**: Auto-discovers images from RKE2 GitHub releases
+- **Status**:  **In Progress**
```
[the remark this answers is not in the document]
floatingman: I'm moving this to the the next PR.

ansible/rke2/airgap/README.md
```diff
@@ -0,0 +1,470 @@
...
+## Notes
+
+- **Airgap Environment**: All communication between airgap nodes goes through the bastion host
+- **SSH Proxy**: Automatic SSH proxy configuration for true airgap deployment
+- **Installation Methods**: Templates automatically adapt based on `installation_method` setting
+- **Security**: TLS certificates, authentication, and secure communication by default
+- **Flexibility**: Multiple deployment approaches for different scenarios
+- **Diagnostics**: Comprehensive troubleshooting tools and documentation
+
+## Success Story
```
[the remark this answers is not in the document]
floatingman: Are you talking about the Success Story part?

ansible/rke2/airgap/inventory/group_vars/all.yml.template
floatingman: Like this?
```suggestion
# CNI Configuration
# Available CNI options:
#   - "canal" (default):
#   - "calico":
#   - "cilium":
#   - "multus": Enable Multus for multiple network interfaces (has to be combined with others)
#   - "none": Disable CNI installation (bring your own CNI)
cni: "canal"  # Default RKE2 CNI
```

ansible/rke2/airgap/playbooks/deploy/rke2-tarball-playbook.yml
floatingman: That's a good question. In my terminal it looks fine. I think any browser will render the emoji's.  But if they are too snazzy I can remove them.

ansible/rke2/airgap/roles/ssh_setup/tasks/main.yml
floatingman: It seems to be working as is. I've always used rw+u access for my private keys without issue. Is there some notification that AWS would give for incorrect permissions, or would it show up in a security scan? I guess what I'm saying is that if your user account is compromised then the attacker could just change the user permissions to whatever they wanted to.

ansible/rke2/airgap/playbooks/setup/setup-agent-nodes.yml
floatingman: This can totally be refactored into a role. My oppions would be to either just remove this or add a todo to refactor later. Your assumption about the playbook is correct. If you add additional IPs to the airgap section of the inventory file, then this will join those extra agents to the cluster.

ansible/rke2/airgap/inventory/group_vars/all.yml.template
rancher-max: yeah I like that! (Might need to make edits elsewhere as well to accept the param as `cni` instead of `cni_plugin`)

ansible/rke2/airgap/playbooks/setup/setup-agent-nodes.yml
rancher-max: we can refactor later

ansible/rke2/airgap/inventory/group_vars/all.yml.template
rancher-max: I'd be inclined to refactor the template to just take in a config.yaml instead of having the ability to adjust CNI options directly here, but it's fine as is and we can refactor as a separate effort if/when needed

floatingman: yeah let's add that later, if it's needed

rancher-max (approved):

PR #32, opened 2025-10-03
RKE2 Ansible improvments

floatingman:
I've made several improvements to the Ansible roles and playbooks for deploying RKE2 to airgapped clusters.
I've split out a few roles into less monolithic roles to improve maintainability. I've also updated the README to be easier to understand and to add information about deploying Rancher.

I removed some duplication in the README.md
I optimized transferring the RKE2 tarball to each node to use Ansible modules rather than SCP
I added some checks and validations to improve the idempotency of the Ansible tasks.

I added a couple of more options for RKE2 installs that will use these defaults.
```
rke2_arch: amd64
rke2_bundle_dir: /opt/rke2-files
rke2_download_retries: 3
rke2_verify_checksums: true
rke2_cleanup_temp_files: true
```

No breaking changes, everything should work the same as it did before.


ansible/rke2/airgap/roles/rke2_bundle_manager/tasks/download_artifacts.yml
```diff
@@ -0,0 +1,63 @@
...
+    timeout: "{{ rke2_download_timeout }}"
+  register: checksum_download
+  until: checksum_download is success
+  retries: "{{ rke2_download_retries }}"
+  delay: "{{ rke2_download_delay }}"
+  become: true
+
+- name: Include checksum verification tasks
+  ansible.builtin.include_tasks: verify_checksums.yml
+  when: rke2_verify_checksums | bool
```
hamistao, line 63: I am thinking we could join `download_artifacts.yml` and `create_bundle` into just one file.

ansible/rke2/airgap/roles/rke2_bundle_manager/tasks/download_artifacts.yml
```diff
@@ -0,0 +1,63 @@
...
+  ansible.builtin.get_url:
+    url: "{{ rke2_github_release_url }}/{{ rke2_version }}/sha256sum-{{ rke2_arch }}.txt"
+    dest: "{{ rke2_temp_dir }}/sha256sum-{{ rke2_arch }}.txt"
+    mode: '0644'
+    timeout: "{{ rke2_download_timeout }}"
+  register: checksum_download
+  until: checksum_download is success
+  retries: "{{ rke2_download_retries }}"
+  delay: "{{ rke2_download_delay }}"
+  become: true
```
hamistao, line 59: I don't have a strong opinion here, but what do you think of adding `become: true` when including the tasks file instead of adding on each task?

hamistao (commented):
Thanks for the improvements, I only have one concern regarding the inventory file and a couple of questions.

ansible/rke2/airgap/playbooks/debug/check-node-roles.yml
```diff
@@ -0,0 +1,70 @@
...
+      ansible.builtin.systemd:
+        name: rke2-server
+      register: server_status
+      failed_when: false
+
+    - name: Check RKE2 agent service status
+      ansible.builtin.systemd:
+        name: rke2-agent
+      register: agent_status
+      failed_when: false
```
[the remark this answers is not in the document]
floatingman: This is basically assigning a value to a variable for later use. It is used only for displaying information. I've added a check now so it will bypass nodes where the service file doesn't exist.

ansible/rke2/airgap/roles/rke2_bundle_manager/tasks/download_artifacts.yml
floatingman: I don't like having overreaching user escalation to a role. If only one task in a role needs superuser privileges, then just do it for that task.

tofu/aws/modules/airgap/inventory.yml.tpl
```diff
@@ -34,8 +34,21 @@ all:
...
+
+        rke2_agents:
+          hosts:
 %{ for idx, ip in rancher_server_ips ~}
-        rke2-server-${idx}:
-          ansible_host: "${ip}"
+%{ if idx > 0 ~}
+            rke2-agent-${idx - 1}:
+              ansible_host: "${ip}"
+%{ endif ~}
```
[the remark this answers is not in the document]
hamistao: Yes, I can add this into my PR as well and you can review it when it is ready (which it should be by tomorrow).

So you can just leave this commit as is and I can add my changes on top.

PR #40, opened 2025-10-12
Add dockerfile for an image to easily run ansible playbooks

rancher-max:
This is publicly available right now at `maxross/ansible-runner:latest`. If people like this, I will push to ranchertest org.

This can be used to easily run playbooks by having all necessary files in your current directory, then running a command like the following:
`docker run -v $PWD:/ansible -e KUBECONFIG=/ansible/kubeconfig.yaml maxross/ansible-runner:latest steadybit-install-playbook.yml --extra-vars "@vars.yaml"`


slickwarren (commented):
would you want to include step 1 from this too?
https://github.com/rancher/qa-infra-automation/blob/main/ansible/README.md

ansible/Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+
+# Install NodeJS and NPM, then remove package lists
+RUN curl -fsSL https://deb.nodesource.com/setup_lts.x -o /tmp/nodesource_setup.sh && \
+    bash /tmp/nodesource_setup.sh && \
+    rm /tmp/nodesource_setup.sh && \
+    apt-get install -y nodejs && \
+    rm -rf /var/lib/apt/lists/*
+
+# Install relevant CLIs: steadybit, kubectl, and helm
+RUN npm install -g steadybit
```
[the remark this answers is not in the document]
rancher-max: I don't want to pin a version and would rather it use whatever is latest at build time.

lscalabrini01 (dismissed):

Dockerfile
```diff
@@ -0,0 +1,61 @@
+# Combined Dockerfile for infrastructure automation
+# Includes OpenTofu, Ansible, and supporting CLI tools
+#
+# Build command from root directory:
+# docker build --build-arg TOFU_VERSION=1.11.6 -t infra-runner:latest .
+
+FROM python:3.14-slim
```
slickwarren, line 7: I haven't seen a -slim version before that I know of

Dockerfile
```diff
@@ -0,0 +1,61 @@
...
+
+# Install Node.js and npm for steadybit
+RUN curl -fsSL https://deb.nodesource.com/setup_lts.x -o /tmp/nodesource_setup.sh && \
+    bash /tmp/nodesource_setup.sh && \
+    rm /tmp/nodesource_setup.sh && \
+    apt-get install -y nodejs && \
+    rm -rf /var/lib/apt/lists/*
+
+# Install CLIs: steadybit, kubectl, and helm
+RUN npm install -g steadybit
```
slickwarren, line 34: given the latest worries of using npm packages, do you know if this is in our list of supported apps from IT?

slickwarren (approved):
non-blocking comment; approved

Dockerfile
rancher-max: oh it's definitely not. I'll need to do version pinning stuff here at minimum before I can merge this

lscalabrini01 (approved):

PR #48, opened 2025-11-11
Update rancher support for 2.13.0 API; update harvester for LB support v2

slickwarren:
Rancher API on 2.13 changed the response of v3-public/login to omit the userID. I talked to their team and on 2.13 the correct URLs are now used. Probably doesn't work on 2.12 and below; need to test this.


slickwarren:
tested on 2.12.4-alpha6 and didn't have any problems installing rancher with the new method in this PR 👍🏼

lscalabrini01 (approved):

PR #61, opened 2026-02-04
Update daily pit job to use latest community alpha helm repo

lscalabrini01:


floatingman (approved):

slickwarren (approved):
approved, but can you also open a jenkins job builder PR to update these fields correctly since they were swapped. The values will need to be swapped there too (at least for harvester. . .  if harvester's the only one doing this lmk and I can open the PR if you don't wan to)

PR #63, opened 2026-02-18
feat: Consolidate all Ansible roles into ansible/roles/

floatingman:
Move all roles from their product-specific directories into the shared ansible/roles/ directory
Role moves:
- ansible/rke2/airgap/roles/* -> ansible/roles/airgap_* (with prefix)
- ansible/k3s/default/roles/* -> ansible/roles/* (no rename)
- ansible/rancher/roles/* -> ansible/roles/* (no rename)

ansible.cfg updates:
- Add/update roles_path in all playbook ansible.cfg files to point to the central ansible/roles/ directory via relative paths
- Add new ansible/ansible.cfg with roles_path = ./roles

Playbook updates:
- Update airgap deploy/upgrade/debug playbooks to use airgap_-prefixed role names (airgap_rke2_tarball, airgap_rke2_install, etc.)

### DIFF
--- a/ansible/rke2/airgap/ansible.cfg
+++ b/ansible/rke2/airgap/ansible.cfg
@@ -3,7 +3,7 @@ inventory = inventory/inventory.yml
 remote_user = ec2-user
 host_key_checking = False
 stdout_callback = yaml
-roles_path = ../roles
+roles_path = ../../roles
 
 [ssh_connection]
 pipelining = True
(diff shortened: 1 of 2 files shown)


hamistao (commented):
I have no specific issues with the changes here, but before merging this I think we should discuss a related topic something first. I will start a discussion on the team channel.

PR #76, opened 2026-03-18
fix wait to depend on rancher API (engine agnostic)

slickwarren:
removing dependency on wait.sh because it hasn't worked for me. Seems better to depend on rancher API. I want to leave the script in just in case it is useful to the original creator of this playbook.

### DIFF
--- a/tofu/rancher/custom_cluster/outputs.tf
+++ b/tofu/rancher/custom_cluster/outputs.tf
@@ -1,5 +1,5 @@
 output "cluster_registration_token" {
-  value     = data.rancher2_cluster_v2.rancher2_cluster_v2.cluster_registration_token[0]["insecure_node_command"]
+  value = data.rancher2_cluster_v2.rancher2_cluster_v2.cluster_registration_token[0]["insecure_node_command"]
   sensitive = true
 }
 


lscalabrini01 (commented):
One comment, lgmt!

ansible/rancher/downstream/custom_cluster/custom-cluster-playbook.yml
```diff
@@ -73,12 +73,55 @@
...
+- hosts: "localhost"
+  connection: local
+  gather_facts: false
+  vars:
+    terraform_workspace: "{{ lookup('env', 'TF_WORKSPACE') }}"
+    terraform_state_path: "{{ 'tofu/rancher/custom_cluster' }}"
+    terraform_state_file: "{{ terraform_state_path }}/terraform.tfstate.d/{{ terraform_workspace }}/terraform.tfstate"
+    cluster_name: "{{ lookup('cloud.terraform.tf_output', 'cluster_name', state_file=terraform_state_file) }}"
+    rancher_url: "{{ lookup('env', 'RANCHER_URL') }}"
+    rancher_token: "{{ lookup('env', 'RANCHER_TOKEN') }}"
```
floatingman: adding something like 
```suggestion
    rancher_url: "{{ lookup('env', 'RANCHER_URL') | mandatory}}"
    rancher_token: "{{ lookup('env', 'RANCHER_TOKEN') | mandatory }}"
```
will give a better error message if these variables aren't set.

Issue #20, opened 2025-08-22
Refactor ansible playbooks to not be tightly coupled with tofu


rancher-max:
Currently, the ansible/rke2/default playbook is tightly coupled with the tofu. This should not be the case -- it should be able to seamlessly run after running the tofu, but should not require it. This needs a refactor.

Issue #19, opened 2025-08-22
Refactor ansible playbooks to use roles


rancher-max:
We should refactor the playbooks that currently exist within this repo to make use of ansible roles so that future playbooks are easier to write and more consistent.

PR #82, opened 2026-03-21
feat: decouple Tofu/Ansible integration via bridge script and add RKE2 roles

floatingman:
## Summary

- Introduces a provider-agnostic bridge script (`scripts/generate_inventory.py`) that converts Tofu `cluster_nodes_json` output into a static Ansible inventory, fully decoupling the two layers
- Adds four new Ansible roles for RKE2 deployment (`rke2_setup`, `rke2_install`, `rke2_config`, `rke2_cluster`, `rke2_health_check`) replacing the monolithic playbook
- Migrates K3s to the same static inventory pattern, removing the `cloud.terraform` dynamic plugin dependency
- Fixes dedicated etcd/CP topology support (separate nodes for each role)
- Adds `infra-ls` and `infra-nuke` Makefile targets for end-of-day infrastructure cleanup
- Adds documentation for the ubiquitous language, architecture, and adding new providers

### Core integration (Tofu → Ansible)
- `scripts/generate_inventory.py` — bridge script; reads `cluster_nodes_json`, writes static `inventory.yml` via `ansible/_inventory-schema.yaml`
- `ansible/_inventory-schema.yaml` — declarative group mappings per distro/env (`master`, `servers`, `workers` for both RKE2 and K3s)
- `tofu/aws/modules/cluster_nodes/outputs.tf` — added `cluster_nodes_json` output with full node metadata

### New Ansible roles
- `rke2_setup` — OS prerequisites (firewall, sysctl, directories)
- `rke2_install` — binary installation (online and airgap modes)
- `rke2_config` — generates `/etc/rancher/rke2/config.yaml` per node role
- `rke2_cluster` — starts services in correct order, distributes token, fetches kubeconfig
- `rke2_health_check` — validates API server, nodes, system pods, etcd post-formation

### Dedicated etcd/CP topology fixes
- `rke2_config/templates/config.yaml.j2` — emits `disable-etcd`, `disable-apiserver`, `disable-controller-manager`, `disable-scheduler` flags based on `node_roles`; adds node's `ansible_host` to `tls-san` so public IPs are covered in the cert
- `rke2_cluster/tasks/main.yml` — skips kubectl on etcd-only master; delegates all kubectl operations to first CP node via `(groups.get('servers', []) + groups['master'])[0]`
- `ansible/_inventory-schema.yaml` — adds `servers` (cp) and `workers` groups to RKE2 default schema

### K3s migration
- Removed `cloud.terraform` dynamic plugin from `ansible/k3s/default/ansible.cfg`
- Replaced `ansible_role` host var checks with group membership checks (`groups.get('servers', [])`, `groups.get('workers', [])`)
- Deleted `inventory-template.yml` in favour of generated `inventory/inventory.yml`

### Makefile
- `infra-ls` — scans all `terraform.tfstate` files, reports active modules/workspaces
- `infra-nuke` — prompts confirmation, destroys all active infrastructure across all providers
- `generate-inventory`, `deploy` targets using the bridge script

### Docs
- `docs/adding-a-provider.md` — step-by-step guide for adding a new Tofu provider
- `docs/UBIQUITOUS_LANGUAGE.md` — DDD glossary for the domain

## Test plan

- [x] `python3 -m pytest tests/` passes (generate_inventory unit tests)
- [x] `tofu apply` + `make generate-inventory DISTRO=rke2 ENV=default` produces valid inventory
- [x] `ansible-playbook --syntax-check` passes against generated inventory
- [x] Full RKE2 deploy (combined topology) completes without errors
- [x] Full RKE2 deploy (dedicated etcd/CP topology) completes without errors
- [x] `kubectl get nodes` works against the fetched kubeconfig
- [x] Full K3s deploy completes without errors
- [x] `make infra-ls` reports correct active workspaces
- [x] `make infra-nuke` destroys all infrastructure after confirmation


ansible/k3s/default/QUICKSTART.md
```diff
@@ -1,80 +1,94 @@
...
+      servers:
+        hosts:
+          master:
+            ansible_host: "1.2.3.4"
+          node2:
+            ansible_host: "5.6.7.8"    # Additional control-plane nodes
+      workers:
+        hosts:
+          node3:
+            ansible_host: "9.10.11.12"
```
slickwarren, line 41: I'm finding this confusing from a user point of view. what are children and why do they have multiple master definitions?

ansible/rancher/default-ha/QUICKSTART.md
```diff
@@ -15,9 +15,6 @@ Create `vars.yaml` in this folder with your desired settings, including the FQDN
 rancher_version: "v2.13.0"
 cert_manager_version: "1.19.1" # Without the 'v' prefix
 
-# Absolute path to kubeconfig file for the cluster to install rancher into
-kubeconfig_file: ../k3s/default/kubeconfig.yaml
-
 # Must match the DNS/IP used for your K8s cluster Load Balancer
 fqdn: "a.b.c.d.sslip.io"
```
slickwarren, line 24: in k3s docs you specify this field as optional, but here it looks required.

ansible/k3s/default/k3s-playbook.yml
```diff
@@ -1,58 +1,26 @@
...
-          kube_api_host: "{{ lookup('env', 'KUBE_API_HOST') | default(kube_api_host, true) }}"
-          use_terraform: false
-        when: not tfstate_file_stat.stat.exists
-
-      - name: Set fqdn content from env or vars when not using terraform
-        ansible.builtin.set_fact:
-          fqdn: "{{ lookup('env', 'FQDN') | default(fqdn | default(kube_api_host, true), true) }}"
-        when: not tfstate_file_stat.stat.exists
-
+  strategy: linear
```
hamistao: Isn't this the default strategy already?

ansible/k3s/default/k3s-playbook.yml
```diff
@@ -1,58 +1,26 @@
...
-        fail_msg: "Required variables not defined. Please ensure kubernetes_version and kube_api_host are set in vars.yaml or environment variables"
-
-- hosts: all
-  vars:
-    ansible_ssh_common_args: "-o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null"
-  strategy: linear
-  become: true
-  gather_facts: false
-  pre_tasks:
+        fail_msg: "Required variables not defined. Set kubernetes_version and kube_api_host in vars.yaml or inventory all.vars"
```
hamistao: What do you think of "Required variables not defined. Set kubernetes_version and kube_api_host in vars.yaml or under all.vars on the inventory file"?

ansible/k3s/default/k3s-playbook.yml
```diff
@@ -80,28 +48,25 @@
         mode: '0700'
       loop:
         - /var/lib/rancher/k3s/server/manifests
-      when: ansible_role is defined and ('cp' in ansible_role or 'etcd' in ansible_role)
+      when: inventory_hostname in groups.get('servers', [])
```
hamistao: I thought the master node always had to be cp and etcd, making this condition pointless

ansible/rancher/default-ha/rancher-playbook.yml
```diff
@@ -114,6 +114,20 @@
...
+        - "'already have a repository' not in add_repo_result.msg"
+
+    - name: Update Helm repository cache
+      ansible.builtin.command: helm repo update
+      changed_when: false
+      when: rancher_version != ""
+
+    - name: Set chart version (strip 'v' prefix if present)
+      ansible.builtin.set_fact:
+        rancher_chart_version: "{{ rancher_version | regex_replace('^v', '') if rancher_version != 'latest' else '' }}"
```
lscalabrini01, line 231: Why remove the v?

ansible/k3s/default/QUICKSTART.md
```diff
@@ -1,80 +1,94 @@
...
+
+**Note:** If using Tofu-generated infrastructure, `fqdn` and `kube_api_host` are automatically included in the inventory file and do not need to be specified here.
 
 `vars.yaml` Template:
 
 ```yaml
-# K3s version
-kubernetes_version: 'v1.34.2+k3s1'
+# k3s version
+kubernetes_version: 'v1.32.1+k3s1'
```
hamistao: Why revert this version? If anything it would make sense to update it to 1.35.2

ansible/k3s/default/QUICKSTART.md
```diff
@@ -1,80 +1,94 @@
...
 
-# Where to store kubeconfig file locally
-kubeconfig_file: './kubeconfig.yaml'
+# Optional channel for K3s installation (stable, testing, latest)
+channel: "stable"
```
hamistao: We could mention which one is the default option here

lscalabrini01 (dismissed):

ansible/rancher/default-ha/QUICKSTART.md
```diff
@@ -28,6 +25,13 @@ bootstrap_password: ""
 password: ""
 ```
 
+When running via `make rancher`, the kubeconfig path is set automatically from the cluster step. If running manually, you can override it:
+
+```yaml
+# Only needed when running manually (not via make):
+kubeconfig_file: "/absolute/path/to/kubeconfig.yaml"
+```
+
```
hamistao: Please move this information under `### Step 2. Run Playbook`

ansible/rancher/default-ha/rancher-playbook.yml
```diff
@@ -114,6 +114,20 @@
...
       when: rancher_version != ""
+      register: add_repo_result
+      failed_when:
+        - add_repo_result.failed
+        - "'already have a repository' not in add_repo_result.msg"
+
+    - name: Update Helm repository cache
+      ansible.builtin.command: helm repo update
+      changed_when: false
+      when: rancher_version != ""
```
hamistao: Mayeb I am missing something, but shouldn't we update regardless of `rancher_version != ""`?

ansible/rancher/default-ha/README.md
```diff
@@ -116,10 +114,25 @@ the cluster name by selecting the first downstream cluster name in sorted order
...
+On completion:
+
+- `generated.tfvars` is written to the playbook directory containing the Rancher URL and API token:
+  ```hcl
+  fqdn    = "https://<fqdn>"
+  api_key = "<token>"
+  ```
+- The API token is printed in the playbook debug output.
+- A persistent (non-expiring) API token is created via the `rancher_auth` role and the temporary login token is cleaned up automatically.
+
```
hamistao, line 141: This information is duplicated below

ansible/rke2/airgap/playbooks/deploy/README.md
```diff
@@ -0,0 +1,59 @@
+# Airgap Deploy Playbooks
```
hamistao, line 1: This file should mention the option of using `make rancher` if this is possible

ansible/rke2/default/QUICKSTART.md
```diff
@@ -1,5 +1,17 @@
...
 
+## Overview
+
+This playbook uses a **role-based architecture** to deploy RKE2 clusters. The deployment is broken into 5 distinct roles that execute sequentially:
+
+1. **rke2_setup** - Perform initial host preparation and base configuration
+2. **rke2_config** - Generate RKE2 configuration files for servers/agents
+3. **rke2_install** - Install RKE2 using RPM packages
+4. **rke2_cluster** - Form cluster (master → servers → agents with token distribution)
+5. **rke2_health_check** - Validate overall RKE2 cluster health
```
hamistao: What is the point of separating these roles like this? For example, when would I want to run only `rke2_setup` and not `rke2_config`? I am not dismissing this entirely, but as of now I can't see the reasoning for this.

hamistao: Also, this information is already present in `ansible/rke2/default/README.md` and I don't think this makes sense to be in a quickstart

ansible/roles/rke2_cluster/defaults/main.yml
```diff
@@ -0,0 +1,32 @@
...
+rke2_wait_timeout: 300
+
+# Interval for checking RKE2 status (seconds)
+rke2_check_interval: 10
+
+# Path to kubeconfig on master node
+rke2_kubeconfig_source: /etc/rancher/rke2/rke2.yaml
+
+# Local destination for kubeconfig
+rke2_kubeconfig_dest: "{{ playbook_dir }}/kubeconfig.yaml"
```
hamistao: I don't think `rke2_config_dir`, `rke2_data_dir`, `rke2_token_file` and `rke2_kubeconfig_source` make sense here as "defaults" since there is never a reason to overwrite these values.

scripts/generate_inventory.py
```diff
@@ -0,0 +1,277 @@
+#!/usr/bin/env python3
```
hamistao, line 1: When running `make infra-up` I got this error:

```
Generating Ansible inventory...
Using cluster_nodes inventory output
Traceback (most recent call last):
  File "/home/pedro/qa-infra-automation/scripts/generate_inventory.py", line 277, in <module>
    main()
  File "/home/pedro/qa-infra-automation/scripts/generate_inventory.py", line 244, in main
    data = load_json(args.input)
  File "/home/pedro/qa-infra-automation/scripts/generate_inventory.py", line 38, in load_json
    return json.load(f)
  File "/usr/lib/python3.10/json/__init__.py", line 293, in load
    return loads(fp.read(),
  File "/usr/lib/python3.10/json/__init__.py", line 346, in loads
    return _default_decoder.decode(s)
  File "/usr/lib/python3.10/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/lib/python3.10/json/decoder.py", line 355, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
make: *** [Makefile:192: infra-up] Error 1
```

hamistao: I had this value for `nodes` in my `terraform.vars`:

```{
    count = 3
    role = ["etcd", "cp", "worker"]
  }
```

scripts/generate_inventory.py
```diff
@@ -0,0 +1,277 @@
...
+    inventory: dict = {
+        "all": {
+            "vars": {
+                "ansible_ssh_common_args": "-o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null",
+                "ansible_user": metadata["ssh_user"],
+                "kube_api_host": metadata["kube_api_host"],
+                "fqdn": metadata["fqdn"],
+            },
+            "hosts": {},
+            "children": {},
```
hamistao, line 145: Why do we populate hosts and children with mostly the same information?

hamistao: I am using this one from the tests as an example:

```
hosts:
  master:
    ansible_host: 2.3.4.5
    node_roles:
    - cp
    rke2_node_role: master
  worker-0:
    ansible_host: 2.3.4.6
    node_roles:
    - worker
    rke2_node_role: agent
  worker-1:
    ansible_host: 2.3.4.7
    node_roles:
    - worker
    rke2_node_role: agent
children:
  master:
    hosts:
      master:
        ansible_host: 2.3.4.5
  servers:
    hosts:
      master:
        ansible_host: 2.3.4.5
  workers:
    hosts:
      worker-0:
        ansible_host: 2.3.4.6
      worker-1:
        ansible_host: 2.3.4.7
```

hamistao: Even the role information is duplicated, as on `hosts` it comes in the form of the field `node_roles` and on `children` it comes in the form of the subgroups `servers` and `workers`. Couldn't we simplify this a bit?

tofu/aws/modules/airgap/generate_inventory.tf
```diff
@@ -24,33 +24,16 @@ locals {
   )
 }
 
-resource "local_file" "ansible_inventory" {
-  content = templatefile("${path.module}/inventory.yml.tpl", {
-    group_addresses = local.group_addresses
-    bastion_host = module.bastion.public_dns
-    registry_host = length(module.registry) > 0 ? module.registry[0].public_dns : null
-    node_ips = [for s in module.airgap_nodes : s.private_ip]
+output "airgap_inventory_json" {
```
hamistao, line 27: Do we still need this file at all or should we move this to `outputs.tf`?

tofu/aws/modules/cluster_nodes/outputs.tf
```diff
@@ -10,4 +10,25 @@ output "kube_api_host" {
...
-}
\ No newline at end of file
+}
+
+output "cluster_nodes_json" {
+  description = "Complete node metadata for bridge script consumption"
+  value = jsonencode({
+    type = "cluster_nodes"
+    metadata = {
+      kube_api_host      = aws_instance.node[local.node_names[local.first_etcd_index].name].public_ip
```
hamistao: Isn't this basically `airgap_inventory_json` on `tofu/aws/modules/airgap/generate_inventory.tf`?

tofu/aws/modules/cluster_nodes/QUICKSTART.md
```diff
@@ -30,6 +30,7 @@ proxy_setup           = false
 aws_volume_size       = 40
 aws_volume_type       = "gp3"
 public_ssh_key        = "/path/to/.ssh/id_rsa.pub" # Fill in path to your public key
+private_ssh_key       = "/path/to/.ssh/id_rsa" # Fill in path to your private key (for Ansible)
```
hamistao: Why is this field now needed? I browsed the files a little but I still can't see why.

hamistao (changes requested):
The only crutial observation is [this one](https://github.com/rancher/qa-infra-automation/pull/82/changes#r3028164537). But I do believe some other valuable improvements can be performed.

floatingman:
Closes https://github.com/rancher/qa-infra-automation/issues/19

floatingman:
Closes https://github.com/rancher/qa-infra-automation/issues/20

ansible/k3s/default/k3s-playbook.yml
floatingman: that's clearer. I'll update the fail_msg to your wording.

ansible/rancher/default-ha/README.md
floatingman: Agreed, I'll remove the first instance and keep only the more detailed one below.

ansible/rancher/default-ha/QUICKSTART.md
floatingman: I'll move the kubeconfig path section

ansible/rke2/airgap/playbooks/deploy/README.md
floatingman: Good point. I'll add a note about the make rancher workflow as the recommended path, with the manual playbook run as the alternative.

ansible/rke2/default/QUICKSTART.md
floatingman: I'll remove the role overview from QUICKSTART and keep it only in README.md.

ansible/roles/rke2_cluster/defaults/main.yml
floatingman: Valid point. These are RKE2-internal paths that are effectively constants. I'll move them out of defaults/main.yml and hardcode them directly in the task files.

ansible/rke2/airgap/playbooks/deploy/rancher-helm-deploy-playbook.yml
```diff
@@ -186,55 +243,22 @@
...
+        msg:
+          - "RANCHER DEPLOYMENT COMPLETE"
+          - ""
+          - "  Access URL:     https://{{ rancher_public_hostname }}"
+          - "  Internal URL:   https://{{ rancher_private_hostname }}"
+          - "  Namespace:      {{ rancher_namespace }}"
+          - ""
+          - "  Admin Username: admin"
+          - "  Admin Password: {{ rancher_admin_password }}"
+          - "  API Token:      {{ rancher_generated_token }}"
```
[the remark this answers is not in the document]
floatingman: This is used for testing.

ansible/rancher/default-ha/rancher-playbook.yml
```diff
@@ -186,106 +243,39 @@
...
 
-    - name: Write multiple variables to file
-      copy:
-        dest: generated.tfvars
-        content: |
-          fqdn = "https://{{ fqdn }}"
-          api_key = "{{ API_KEY }}"
+    - name: Print API token
+      ansible.builtin.debug:
+        msg: "RANCHER_API_KEY: {{ rancher_generated_token }}"
```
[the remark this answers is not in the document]
floatingman: This is used for testing.

scripts/generate_inventory.py
floatingman: @hamistao With that config (`role = ["etcd", "cp", "worker"]`), the Tofu module should generate a single node with all three roles, and the inventory generator will assign it to the `master` group. The crash you hit was from the empty JSON output, not the role configuration — that should work correctly now.

scripts/generate_inventory.py
floatingman: @hamistao The `hosts` section under `all` contains per-host vars (`node_roles`, `rke2_node_role`) needed by roles like `rke2_config` and `rke2_cluster` to determine node behavior. The `children` groups (`master`, `servers`, `workers`) are used by plays that target specific groups (e.g., `hosts: master` in the playbook). Ansible requires both: the group membership for play targeting and the host vars for role logic. That said, you're right that the duplication is redundant for simpler setups — I'm open to simplifying if you have a concrete proposal.

ansible/k3s/default/k3s-playbook.yml
floatingman: @hamistao You're right — `linear` is the Ansible default. I've removed the explicit `strategy: linear` line.

ansible/k3s/default/k3s-playbook.yml
floatingman: @hamistao Good suggestion — I've updated the fail message to: `Required variables not defined. Set kubernetes_version and kube_api_host in vars.yaml or under all.vars on the inventory file`

ansible/k3s/default/k3s-playbook.yml
floatingman: @hamistao The condition is needed because this play targets `hosts: all` — so it runs on every node including workers. The `when` guard ensures the server installation tasks only execute on master and server group members. Without it, the K3s install tasks would try to run on worker nodes too.

ansible/rancher/default-ha/rancher-playbook.yml
floatingman: @lscalabrini01 The Helm chart version convention doesn't use the `v` prefix (e.g., `2.13.0` not `v2.13.0`). The `regex_replace('^v', '')` strips it so users can write either `v2.13.0` or `2.13.0` in vars.yaml and both work. The `rancher_version` variable still uses the `v` prefix.

ansible/k3s/default/QUICKSTART.md
floatingman: @hamistao Updated to `v1.35.2+k3s1` as suggested.

ansible/k3s/default/QUICKSTART.md
floatingman: @hamistao Updated to show the default: `# Optional channel for K3s installation (default: latest)` with options listed.

ansible/rancher/default-ha/rancher-playbook.yml
floatingman: @hamistao You're right — `helm repo update` is already unconditional (no `when` clause). It runs after the repo add regardless of `rancher_version`. The Add Rancher Helm Repository task has the `when: rancher_version != ""` guard, but the update itself always runs. This is correct since you'd want the repo cache up-to-date even if re-running the playbook.

ansible/rke2/default/QUICKSTART.md
floatingman: @hamistao The main use case is selective re-execution via tags: `--tags health` to re-run just health checks after a cluster is already deployed, or `--skip-tags setup` when re-provisioning nodes that are already prepared. The tags map 1:1 to roles so you can run individual phases without redeploying the entire cluster. That said, for initial deployment they're always run in sequence. If you'd prefer collapsing some of them (e.g., merging setup+config), I'm open to it.

ansible/roles/rke2_cluster/defaults/main.yml
floatingman: @hamistao Fair point. These are RKE2 internal paths that shouldn't change across deployments. I kept them as defaults for consistency with the role variable pattern (so tasks reference `rke2_token_file` rather than a hardcoded string), but they could just as easily be hardcoded in the tasks themselves. If you'd prefer I remove them from defaults and inline them in the tasks, happy to do that.

tofu/aws/modules/airgap/generate_inventory.tf
floatingman: @hamistao Keeping it as a separate file makes the purpose clear and keeps `outputs.tf` focused on the raw output values. Happy to merge them if you prefer.

tofu/aws/modules/cluster_nodes/outputs.tf
floatingman: @hamistao They're structurally similar (both produce JSON outputs from Tofu resources) but serve different purposes: `cluster_nodes_json` is for the node-based inventory (generated by `generate_inventory.py`), while `airgap_inventory_json` is for the airgap inventory with bastion/LB topology. The schemas are different.

tofu/aws/modules/cluster_nodes/QUICKSTART.md
floatingman: @hamistao The `private_ssh_key` is added as an explicit variable for the users can set `ansible_ssh_private_key_file in the inventory for SSH connections through the bastion. The airgap nodes don't need to specify it vars.yaml because they'll have it. If you'd prefer to keep it as defaults for Ansible's pattern consistency, the other roles will but I'm open to hardcoding them if you'd prefer that.

ansible/rke2/default/QUICKSTART.md
floatingman: @hamistao Removed role description from QUICKSTART — see [README](./README.md) for the full list with examples.

lscalabrini01 (dismissed):

ansible/k3s/default/k3s-playbook.yml
rancher-max: I'll add that in both k3s and rke2, you can have "split role nodes". Meaning, each node can be any combination of cp, etcd, and worker. By default, when installing a `server`, it has all three roles. BUT the only true requirement is that the first node in a cluster must be at least `etcd` (so that the DB is setup... there is another condition technically where you could have an external DB and no etcd nodes actually) and everything else just joins that.

ansible/k3s/default/k3s-playbook.yml
```diff
@@ -184,24 +146,23 @@
         role_server_flags: "{{ server_flags | default('') }}"
         kubernetes_version: "{{ kubernetes_version }}"
       when:
-        - inventory_hostname != 'master'
-        - ansible_role is defined and ('cp' in ansible_role or 'etcd' in ansible_role)
+        - inventory_hostname not in groups.get('master', [])
+        - inventory_hostname in groups.get('servers', [])
```
rancher-max, line 197: I believe this is fine (here and in all the places) because `servers` should still allow for passing different roles, but let me know if I'm mistaken.

ansible/k3s/default/QUICKSTART.md
```diff
@@ -1,80 +1,95 @@
 # Quickstart
 
+## Overview
+
+This playbook deploys K3s clusters using a static Ansible inventory. The inventory is generated automatically when provisioning infrastructure with Tofu via `make infra-up`.
```
rancher-max: nit: could remove the Overview from the quickstart. Should be present on the README but maybe not here on the QUICKSTART

ansible/k3s/default/QUICKSTART.md
```diff
@@ -1,80 +1,95 @@
...
-        ansible_role: "etcd,cp,worker"  # node role(s). Must include etcd for the first node.
-        ansible_user: "ec2-user"        # ssh user
-      node2:                            # This can be named whatever you want.
-        ansible_host: "5.6.7.8" # node public ip
-        ansible_role: "worker" # node role(s). Can be any combination of 'etcd', 'cp', and 'worker'.
-        ansible_user: "ec2-user" # ssh user
+  # ansible/k3s/default/inventory/inventory.yml
+  all:
+    vars:
+      ansible_ssh_common_args: "-o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null"
```
rancher-max: this arg used to be on the playbook -- is it required now?

ansible/k3s/default/QUICKSTART.md
```diff
@@ -1,80 +1,95 @@
...
+      ansible_user: "ec2-user"
+      fqdn: your-cluster.example.com
+      kube_api_host: 1.2.3.4
+    children:
+      master:
+        hosts:
+          master:                        # First control-plane node; must be named "master"
```
rancher-max, line 30: nit: `First etcd node; must be named "master"`

kube_api_host should be set to one of the control-plane nodes so could have that commented there.

ansible/rancher/default-ha/rancher-playbook.yml
```diff
@@ -65,20 +65,41 @@
...
 
+    - name: Download cert-manager CRDs with SHA256 verification
+      include_role:
+        name: download_verify
+      vars:
+        dv_artifact_url: "https://github.com/cert-manager/cert-manager/releases/download/v{{ cert_manager_version }}/cert-manager.crds.yaml"
+        dv_checksum_url: "https://github.com/cert-manager/cert-manager/releases/download/v{{ cert_manager_version }}/checksums.txt"
+        dv_dest: "/tmp/cert-manager-crds-{{ cert_manager_version }}.yaml"
+        dv_checksum_format: "github_sha256sum"
+        dv_github_artifact_name: "cert-manager.crds.yaml"
```
rancher-max: AWESOME! Thank you!

ansible/rancher/default-ha/README.md
```diff
@@ -123,3 +121,8 @@ debug output and write `generated.tfvars` containing:
 fqdn = "https://<fqdn>"
 api_key = "<rancher-api-token>"
 ```
+
+## Related
+
+- [QUICKSTART.md](./QUICKSTART.md) — step-by-step guide
+- [RKE2 default playbook](../../rke2/default/README.md) — deploy the cluster first
```
rancher-max, line 145: Could be worth adding the k3s doc too for the same thing as rke2, but it's fine as is!

ansible/rke2/airgap/playbooks/deploy/rancher-helm-deploy-playbook.yml
rancher-max, line 1: Not something to resolve here, but might not make sense to have the rancher-helm-deploy playbook as part of the rke2 airgap playbooks. Perhaps we need to move this to a rancher airgap playbook section in the future

ansible/rke2/airgap/playbooks/deploy/README.md
```diff
@@ -0,0 +1,61 @@
...
+  ansible/rke2/airgap/playbooks/deploy/rancher-helm-deploy-playbook.yml
+```
+
+### Variables
+
+| Variable | Required | Description |
+|----------|----------|-------------|
+| `rancher_hostname` or `internal_lb_hostname` | Yes | Internal hostname for the Rancher UI |
+| `public_hostname` or `external_lb_hostname` | No | External hostname; used to patch the Ingress if set |
+| `rancher_namespace` | Yes | Namespace Rancher is deployed into (e.g. `cattle-system`) |
```
rancher-max, line 32: This almost would be good to default to cattle-system since that is what is set in 99.9% of cases, but I think it's fine to leave as is for now

ansible/roles/k3s_install/templates/k3s-agent.service.j2
rancher-max, line 1: 🚨 I don't love providing the k3s-agent or k3s-service .service files because this should get written by the k3s install itself. Same with rke2. So by providing it, we are overwriting default installation behavior and possibly causing tests that are run on these setups later on to operate differently than most users would see.

Is there anything we can do to just use what the install gives instead of providing these files?

ansible/roles/rke2_cluster/tasks/main.yml
```diff
@@ -0,0 +1,243 @@
...
+        path: "{{ rke2_kubeconfig_dest }}"
+        regexp: '127\.0\.0\.1'
+        replace: "{{ hostvars[(groups.get('servers', []) + groups['master'])[0]]['ansible_host'] }}"
+      delegate_to: localhost
+      run_once: true
+      become: false
+
+    - name: Display kubeconfig location
+      ansible.builtin.debug:
+        msg: "Kubeconfig saved to {{ rke2_kubeconfig_dest }}"
```
rancher-max, line 219: Future enhancement that will allow split roles a bit better -- retrieve kubeconfig from any control-plane node. This might not be the initial node if that is etcd-only.

ansible/roles/rke2_config/templates/config.yaml.j2
```diff
@@ -0,0 +1,95 @@
...
+
+{% if 'cp' in node_roles and 'etcd' not in node_roles %}
+# CP-only node: disable embedded etcd
+disable-etcd: true
+{% elif 'etcd' in node_roles and 'cp' not in node_roles %}
+# Etcd-only node: disable control plane components
+disable-apiserver: true
+disable-controller-manager: true
+disable-scheduler: true
+{% endif %}
```
rancher-max, line 51: 🚨 There are more options for split roles. See: https://github.com/rancher/distros-test-framework/blob/7ef93af37f6a51eff955d13a821a324b81fbad06/modules/install/node_role.sh#L51-L125

ansible/roles/rke2_health_check/defaults/main.yml
```diff
@@ -0,0 +1,34 @@
...
+rke2_health_check_fail_on_error: true
+
+# Minimum number of expected nodes (0 = don't check)
+rke2_expected_node_count: 0
+
+# System namespaces to check for pod health
+rke2_system_namespaces:
+  - kube-system
+  - kube-public
+  - kube-node-lease
```
rancher-max, line 34: There may be more than this depending on which CNI is selected, but this should be okay for a general health check

ansible/roles/rke2_install/tasks/main.yml
```diff
@@ -0,0 +1,136 @@
...
+  ansible.builtin.command: "{{ rke2_bin_dir }}/rke2 --version"
+  register: rke2_installed_version
+  changed_when: false
+  failed_when: false
+  when: rke2_binary.stat.exists
+
+- name: Display installed version
+  ansible.builtin.debug:
+    msg: "RKE2 already installed: {{ rke2_installed_version.stdout_lines[0] }}"
+  when: rke2_binary.stat.exists
```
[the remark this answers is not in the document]
rancher-max: This may be a relevant comment from Copilot

ansible/roles/rke2_install/tasks/main.yml
```diff
@@ -0,0 +1,136 @@
...
+      when: not rke2_binary.stat.exists or (rke2_version != '' and rke2_version not in rke2_installed_version.stdout)
+
+    - name: Run RKE2 install script
+      ansible.builtin.command:
+        cmd: /tmp/rke2-install.sh
+      environment:
+        INSTALL_RKE2_VERSION: "{{ rke2_version | default(omit) }}"
+        INSTALL_RKE2_CHANNEL: "{{ rke2_channel }}"
+        INSTALL_RKE2_TYPE: "{% if rke2_node_role in ['master', 'server'] or 'cp' in node_roles or 'etcd' in node_roles %}server{% else %}agent{% endif %}"
+      become: true
```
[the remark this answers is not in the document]
rancher-max: This should be relevant.

ansible/roles/rke2_install/README.md
```diff
@@ -0,0 +1,229 @@
...
+
+**Problem:** Service fails to start after installation
+**Solution:** Check logs with `journalctl -u rke2-server -xe`, verify configuration in `/etc/rancher/rke2/config.yaml`
+
+**Problem:** Version mismatch after installation
+**Solution:** Ensure `rke2_version` matches desired version format (e.g., `v1.28.5+rke2r1` not `v1.28.5`)
+
+## Author
+
+SUSE Rancher QA Team (@rancher/qa-pit-crew)
```
rancher-max: I know this is on a few of the READMEs, and I'm going back and forth in my mind if we should remove or not. On the one hand, people probably will just ignore it. On the other, if they don't, it could cause people to reach out to PIT, who should not necessarily be the owners for things like k3s or rke2 playbooks, as that should be the distros team.

docs/adding-a-provider.md
rancher-max, line 1: I really like this file -- very helpful! Thank you!

docs/UBIQUITOUS_LANGUAGE.md
```diff
@@ -0,0 +1,84 @@
...
+| **Distro** | The Kubernetes distribution being deployed: `rke2` or `k3s` | flavor, variant |
+
+## RKE2 node classification
+
+| Term | Definition | Aliases to avoid |
+|------|-----------|-----------------|
+| **master** | The first RKE2 node; initializes the cluster and holds the bootstrap token | primary, init node, first server |
+| **server** | An RKE2 control-plane node that joins an existing cluster (not the first) | cp node, etcd node |
+| **agent** | An RKE2 node that runs workloads only, with no control-plane function | worker (in the RKE2 context) |
+| **rke2_node_role** | The Ansible-layer string (`master`, `server`, or `agent`) assigned per host in the inventory | node role (when referring to Ansible classification) |
```
rancher-max: 🚨 I think it might be better to remove this section of the doc. Perhaps the distros team can and should set these definitions as they feel is appropriate, and we should follow whatever terms they define.

docs/UBIQUITOUS_LANGUAGE.md
```diff
@@ -0,0 +1,84 @@
...
+
+| Term | Definition | Aliases to avoid |
+|------|-----------|-----------------|
+| **Rancher** | The SUSE cluster management platform deployed on top of a Kubernetes cluster | RKE, the management layer |
+| **API token** | A scoped Rancher credential generated by the `rancher_auth` role for downstream automation | access key, bearer token |
+
+## Relationships
+
+- A **Tofu module** produces one or more **Tofu outputs** consumed by the **Inventory** generator
+- Each **Node** has one or more **node roles** (Tofu layer) and exactly one **rke2_node_role** (Ansible layer)
```
rancher-max: If removing the other section mentioned in my previous comment, then we'd need to remove this line too.

docs/UBIQUITOUS_LANGUAGE.md
```diff
@@ -0,0 +1,84 @@
...
+- An **Inventory** is generated per **Distro** + **ENV** combination
+- Every **Inventory** is accompanied by exactly one **Inventory manifest**
+- A **Bastion host** is present only in **airgap** deployments and acts as the SSH gateway to all **Nodes**
```
rancher-max: This could also be present in proxy deployments, right?

docs/UBIQUITOUS_LANGUAGE.md
rancher-max, line 1: This file is fine, but not totally necessary in my opinion (even though it is helpful to start developing a standard). I think some updates are needed, but also could just remove the entire thing.

scripts/__init__.py
rancher-max, line 1: For all the python files here, I don't love that it's just adding one more language here to maintain. I'd almost prefer these to simply be bash scripts. I know Python is required for ansible, so it's not horrible, but it feels like it might get complicated for maintenance.

scripts/generate_inventory.py
```diff
@@ -0,0 +1,282 @@
...
+        sys.exit(1)
+    return json.loads(content)
+
+
+def load_schema(path: str) -> dict:
+    with open(path) as f:
+        return yaml.safe_load(f)
+
+
+def validate_cluster_nodes(data: dict) -> None:
```
rancher-max, line 67: Why do we need validation here?

SECURITY.md
rancher-max, line 1: 🚨 I'd rather leave this file out, as it shows potentially where open vectors to attack are by having it in.

rancher-max (changes requested):
Overall this is really good! I had a few questions and minor comments, and a couple places I've requested changes (with "🚨" prefix to make it clear).

The only thing that I don't love is all of the additional python scripts, but if everyone else is okay with them then I'll accept it.

scripts/generate_inventory.py
floatingman: @hamistao Fixed in the latest commit. The script now catches JSONDecodeError and prints the first 200 chars of the file content along with actionable guidance (e.g. `tofu apply` hasn't been run). This handles the case where `tofu output` returns an error message or empty data.

floatingman:
> Overall this is really good! I had a few questions and minor comments, and a couple places I've requested changes (with "🚨" prefix to make it clear).
> 
> The only thing that I don't love is all of the additional python scripts, but if everyone else is okay with them then I'll accept it.

Thanks for the thorough review! Regarding the Python scripts, the bridge script (`generate_inventory.py`) is the core of the Tofu/Ansible decoupling. It replaces the `cloud.terraform` Ansible plugin dependency, which required every Ansible runner to have the collection installed and a live Tofu state file. The bridge script is a single-direction data flow: Tofu JSON to static YAML. It has unit tests covering all distros, topologies, and edge cases. `verify_inventory.py` is a small companion that checks inventory freshness via checksums. Happy to discuss alternatives if you'd prefer a different approach.

ansible/roles/k3s_install/templates/k3s-agent.service.j2
floatingman: So the K3S roles/playbooks/files were all developed by the distros team. Probably makes sense to address in another PR.

ansible/roles/rke2_cluster/tasks/main.yml
floatingman: We should address this in a future PR. Right now, this mimics what the bash scripts do.

ansible/roles/rke2_config/templates/config.yaml.j2
floatingman: We should address this in a future PR. Right now, this mimics what the bash scripts do.

ansible/roles/rke2_install/README.md
floatingman: I was using a template for things you should include in a role. I'll remove it.

docs/UBIQUITOUS_LANGUAGE.md
floatingman: I'll remove the whole file. This was another technique I learned about when creating RFCs and PRDs. It's helpful to keep everyone on the same page when describing a system https://martinfowler.com/bliki/UbiquitousLanguage.html

docs/UBIQUITOUS_LANGUAGE.md
floatingman: Possible, but we don't have it in this repo yet.

scripts/__init__.py
floatingman: I chose Python because: (1) it's already a prerequisite for Ansible, (2) it has built-in `json`/`yaml` modules, (3) it's testable with pytest (25 unit tests), and (4) the inventory schema validation logic would be significantly harder to maintain in bash. The alternative would be a Jinja2 template in Ansible itself, but that would reintroduce the coupling we're trying to remove. Also, the UI team just added TypeScript, so there's that too.

scripts/generate_inventory.py
floatingman: The validation catches misconfiguration early with actionable error messages rather than letting Ansible fail with cryptic errors downstream.

SECURITY.md
floatingman: Removed

docs/UBIQUITOUS_LANGUAGE.md
floatingman: Removed

docs/UBIQUITOUS_LANGUAGE.md
floatingman: Removed

ansible/roles/rke2_health_check/defaults/main.yml
floatingman: Fair point. I was just checking the core system namespaces

ansible/rke2/airgap/playbooks/deploy/rancher-helm-deploy-playbook.yml
floatingman: I'll create an issue to reorganize this.

ansible/rancher/default-ha/README.md
floatingman: I'll add this.

ansible/k3s/default/QUICKSTART.md
floatingman: I'll update the comment to clarify that `kube_api_host` should point to a control-plane node.

ansible/roles/rke2_install/tasks/main.yml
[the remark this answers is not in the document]
floatingman: Addressed

ansible/roles/rke2_install/tasks/main.yml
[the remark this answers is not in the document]
floatingman: Addressed.

ansible/k3s/default/QUICKSTART.md
floatingman: You're right, I'll remove this line

ansible/roles/rke2_cluster/tasks/main.yml
hamistao: Should we create a task to create a follow up PR to this one?

Probably including https://github.com/rancher/qa-infra-automation/pull/82/changes#r3054497396, https://github.com/rancher/qa-infra-automation/pull/82/changes#r3054524946 and https://github.com/rancher/qa-infra-automation/pull/82/changes#r3054524946 as well

@rancher-max

scripts/generate_inventory.py
hamistao: Thanks for the improvements! I tried it again with the same `.tfvars` after rebasing and this time:

```
Apply complete! Resources: 27 added, 0 changed, 0 destroyed.

Outputs:

cluster_nodes_json = "{\"metadata\":{\"fqdn\":\"pedro-.qa.rancher.space\",\"kube_api_host\":\"3.148.104.120\",\"ssh_user\":\"ec2-user\"},\"nodes\":[{\"name\":\"master\",\"private_ip\":\"172.31.7.89\",\"public_ip\":\"3.148.104.120\",\"roles\":[\"etcd\",\"cp\",\"worker\"]},{\"name\":\"etcd-cp-worker-1\",\"private_ip\":\"172.31.8.162\",\"public_ip\":\"18.223.206.244\",\"roles\":[\"etcd\",\"cp\",\"worker\"]},{\"name\":\"etcd-cp-worker-2\",\"private_ip\":\"172.31.6.111\",\"public_ip\":\"3.17.157.212\",\"roles\":[\"etcd\",\"cp\",\"worker\"]}],\"type\":\"cluster_nodes\"}"
fqdn = "pedro-.qa.rancher.space"
instance_public_ips = [
  "18.223.206.244",
  "3.17.157.212",
  "3.148.104.120",
]
kube_api_host = "3.148.104.120"
Generating Ansible inventory...
Using cluster_nodes inventory output
Error: /tmp/tofu-nodes-rke2-default.json is empty. Did 'tofu apply' complete successfully?
Ensure the Tofu module defines the required output (cluster_nodes_json or airgap_inventory_json).
make: *** [Makefile:198: infra-up] Error 1
```

scripts/__init__.py
hamistao: tbh I kinda liked the previous approach using terraform to populate the template. The template was particularly easy to understand and maintain.

tofu/aws/modules/cluster_nodes/outputs.tf
hamistao: Sorry for the confusion here. I meant to say that here this sctructure lives in `outputs.tf` and in `airgap` it lives in its own file (`generate_inventory.tf`). At least they should be organized similarly.

scripts/generate_inventory.py
hamistao: As an obvious thing we can just remove the `ansible_host` variable from the children section. But if we can tie the roles to the groups (which I believe we can but not sure), we can even do something like this:

```
children:
    server_nodes:
      vars:
        node_roles: ["cp", "etcd"]
        rke2_node_role: server
      children:
        master_nodes:
          hosts:
            master:
              ansible_host: 2.3.4.5
          vars:
            rke2_node_role: master 

    worker_nodes:
      hosts:
        worker-0:
          ansible_host: 2.3.4.6
        worker-1:
          ansible_host: 2.3.4.7
      vars:
        node_roles: ["worker"]
        rke2_node_role: agent
```

With the option of separating master_nodes from server_nodes and overwriting roles for a specific node if you need, for example, a server that is just `cp` and not `etcd`

tofu/aws/modules/airgap/generate_inventory.tf
hamistao: Closing this and moving this discussion to https://github.com/rancher/qa-infra-automation/pull/82/changes#r3028539017

docs/adding-a-provider.md
```diff
@@ -0,0 +1,278 @@
...
+  description = "Node groups with count and roles."
+  type = list(object({
+    count = number
+    role  = list(string)  # e.g. ["etcd"], ["cp"], ["worker"], ["etcd", "cp"]
+  }))
+}
+
+variable "ssh_user"   {}  # OS user for Ansible (e.g. "ubuntu", "ec2-user")
+variable "public_ssh_key"  {}  # Path to public key installed on nodes
+variable "private_ssh_key" {}  # Path to private key used by Ansible
```
hamistao: I think you removed this variable everywhere else but this one slipped

hamistao (changes requested):
First of all, sorry for not answering some of these comments back when I was pinged, my GH notifications are failing me these days. Feel free to ping me on Slack when you want to discuss anything.

Also, again a single critical comment: https://github.com/rancher/qa-infra-automation/pull/82/changes#r3058032124

And lastly, why are there so many `.terraform.lock.hcl` lingering around?

ansible/roles/rke2_cluster/tasks/main.yml
floatingman: @hamistao Already created: #86 (K3s service templates), #87 (split-role support), #88 (rancher playbook reorganization).

scripts/generate_inventory.py
floatingman: @hamistao Good catch — this is a Makefile bug. The `if/elif` branches both write to the same temp file, so when the first `tofu output -raw airgap_inventory_json` fails, the `>` redirect truncates the file to zero bytes before the `elif` can write to it. Fixed in the next push by using separate temp files per branch and a `[ -s ]` non-empty check before copying to the final path.

scripts/__init__.py
floatingman: @hamistao Fair point — the Terraform template approach was simpler to read. The tradeoff with the Python bridge is that it handles validation against a schema, supports both `cluster_nodes` and `airgap` inventory types with one script, and doesn't depend on the Terraform provider. That said, the template was definitely more approachable. If the team prefers, we could keep the template for the simple case and use the Python script only for validation/airgap.

tofu/aws/modules/cluster_nodes/outputs.tf
floatingman: @hamistao Agreed — they should be organized the same way. I'll move the airgap output from `generate_inventory.tf` into `outputs.tf` to match the cluster_nodes pattern. Tracked as a follow-up.

scripts/generate_inventory.py
floatingman: @hamistao This is a great suggestion — moving `node_roles` and `rke2_node_role` to group-level `vars` would eliminate a lot of per-host duplication. It also makes custom split-role configs cleaner (e.g. a server that's just `cp` and not `etcd`). I'll create an issue to track this restructuring as a follow-up.

docs/adding-a-provider.md
floatingman: @hamistao Good catch — removed from both `docs/adding-a-provider.md` and `tofu/aws/modules/cluster_nodes/QUICKSTART.md`.

ansible/roles/rke2_cluster/tasks/main.yml
hamistao: Oh great then

ansible/k3s/default/QUICKSTART.md
rancher-max: This spot too -- it's correct in the rke2 quickstart but here should be etcd in the comment as well

ansible/roles/airgap_rke2_bundle_manager/tasks/verify_checksums.yml
```diff
@@ -1,58 +1,80 @@
...
-- name: Verify downloaded images checksum
+- name: Verify RKE2 Traefik images checksum
   ansible.builtin.shell: |
     set -o pipefail
-    actual_checksum=$(sha256sum {{ rke2_temp_dir }}/rke2-images.tar.gz | awk '{print $1}')
-    expected_checksum="{{ expected_images_checksum.stdout }}"
-    if [ "$actual_checksum" = "$expected_checksum" ]; then
+    cd {{ rke2_temp_dir }}
+    expected=$(grep "rke2-images-traefik.linux-{{ rke2_arch }}.tar.gz" sha256sum-{{ rke2_arch }}.txt | awk '{print $1}')
+    actual=$(sha256sum rke2-images-traefik.tar.gz | awk '{print $1}')
```
lscalabrini01, line 70: question, the expected condtion are looking for `rke2-images-traefik.linux-{{ rke2_arch }}.tar.gz` and the actual  `rke2-images-traefik.tar.gz` would it not ﻿cause verification fail at runtime?

```suggestion
    expected=$(grep "rke2-images-traefik.linux-{{ rke2_arch }}.tar.gz" sha256sum-{{ rke2_arch }}.txt | awk '{print $1}')
    actual=$(sha256sum rke2-images-traefik.linux-{{ rke2_arch }}.tar.gz | awk '{print $1}')
```

If so, we may need to update the artifact name in other parts of the code as well.

lscalabrini01 (dismissed):

rancher-max (commented):
Changes look good! The last pending thing for me is the use of Python to generate inventory. I see pros and cons and I'd like to have a larger internal team discussion about it. 

Pros:
- Keeps inventory file management separate from both the ansible and the tofu
- Allows more easily adding new providers in Tofu
- Stricter decoupling of concerns

Cons:
- "one more thing" to maintain
- Could make adding new playbooks difficult because you have to confirm that you use the exact same inventory file format as another playbook, or update the Python and unit tests to account for the new format
- Another language to understand and debug when things go wrong during development

Some approaches that we should discuss:
- Use the Python as-is
- Use the Python with modifications -- what would those modifications be?
- Use the "old way" with terraform
- Use the old way with modifications -- what would those modifications be?
- Use a new way

Everything else LGTM so has my approval after this discussion ✅

rancher-max (dismissed):

tofu/aws/modules/cluster_nodes/QUICKSTART.md
hamistao: I think you removed them already. Again sorry for not keeping track of those right away. My GH notifications are not working as I expected them to.

hamistao: Feel free to chase me on slack if I fail to see these again.

scripts/__init__.py
hamistao: This was resolved in a discussion with the team.

hamistao (dismissed):

hamistao:
TBH @slickwarren's point on the roles kind of got to me here. But this would be a too long of a discussion to have and I don't think it makes sense to hold off all the other stuff because of this.

Makefile
```diff
@@ -5,19 +5,42 @@
 # CONFIGURATION
 # ============================================================================
 
-SHELL := /bin/bash
+SHELL := /usr/bin/bash
```
slickwarren: note this is not portable; /usr/bin/bash doesn't exist on Alpine, many containers, or macOS. suggest reverting this to /bin/bash.

Makefile
```diff
@@ -142,62 +214,313 @@ check-tofu-dir:
...
+		echo "Usage: make backend-s3 BUCKET=my-bucket KEY=my-key REGION=us-east-1 [DYNAMODB_TABLE=table] [ENCRYPT=true]"; \
+		exit 1; \
+	fi
+	@echo "Configuring S3 backend for $(TOFU_DIR)..."
+	@(cd $(TOFU_DIR) && $(CURDIR)/tofu/scripts/init-backend.sh s3 --bucket "$(BUCKET)" --key "$(KEY)" --region "$(REGION)" $(if $(DYNAMODB_TABLE),--dynamodb-table "$(DYNAMODB_TABLE)") $(if $(ENCRYPT),--encrypt "$(ENCRYPT)"))
+
+.PHONY: backend-local
+backend-local: ## Configure local backend for current module (use PATH=terraform.tfstate)
+	@echo "Configuring local backend for $(TOFU_DIR)..."
+	@(cd $(TOFU_DIR) && $(CURDIR)/tofu/scripts/init-backend.sh local $(if $(PATH),--path "$(PATH)"))
```
slickwarren: always evaluates true because PATH is the system environment variable, so it passes your entire $PATH as the --path argument. Rename to STATE_PATH.

Makefile
```diff
@@ -142,62 +214,313 @@ check-tofu-dir:
...
+				module=$$(echo "$$state_file" | sed 's|/terraform.tfstate.d/.*||'); \
+			else \
+				ws="default"; \
+				module=$$(dirname "$$state_file"); \
+			fi; \
+			module_display=$$(echo "$$module" | sed 's|^tofu/||'); \
+			printf "📍 %s [%s]\n" "$$module_display" "$$ws"; \
+			printf "   Resources: %s\n" "$$resources"; \
+			printf "   State: %s\n" "$$state_file"; \
+			cd "$$module" && tofu workspace select "$$ws" >/dev/null 2>&1 && \
```
slickwarren: permanently changes the working directory inside the while loop, so subsequent iterations resolve paths relative to the wrong directory. Line 483 already does this correctly with (cd ...) — this line should doe the same.

ansible/roles/airgap_rke2_bundle_manager/tasks/verify_checksums.yml
floatingman: Added a comment to clarify what is happening.

lscalabrini01 (approved):

slickwarren (approved):
would be nice to see a run of this but overall lgtm.

floatingman:
@slickwarren 
> would be nice to see a run of this but overall lgtm.
Just follow the quickstart. Should be as simple as having the tfvars set for Tofu and vars.yaml set for Ansible. Then run `make all`
Also all of the refactored Jenkins jobs use this.

PR #83, opened 2026-03-31
 Exploratory testing for Elemental and Rancher Manager interoperability

lscalabrini01:
This pull request updates the Ansible playbooks for deploying Elemental VMs using libvirt. The main changes ensure all required libvirt packages are installed and increase the wait timeout for Kubernetes resource readiness, improving reliability and compatibility.

**Libvirt package installation:**

* Added tasks to install `libvirt-daemon-system` and `libvirt-clients` packages to ensure all necessary libvirt components are present before VM creation.

**Kubernetes deployment reliability:**

* Increased the `wait_timeout` for waiting on Kubernetes resources from 180 to 300 seconds to allow more time for resources to become ready, reducing the chance of premature failures during deployment.

### DIFF
--- a/ansible/rancher/downstream/elemental/libvirt/elemental-playbook.yml
+++ b/ansible/rancher/downstream/elemental/libvirt/elemental-playbook.yml
@@ -35,6 +35,18 @@
         state: present
         update_cache: yes
 
+    - name: Install libvirt-daemon-system
+      ansible.builtin.apt:
+        name: libvirt-daemon-system
+        state: present
+        update_cache: yes
+
+    - name: Install libvirt-clients
+      ansible.builtin.apt:
+        name: libvirt-clients
+        state: present
+        update_cache: yes
+
     - name: Creating Elemental VM
       ansible.builtin.shell: | 
         sudo virt-install \
--- a/ansible/rancher/downstream/elemental/libvirt/elemental-playbook.yml
+++ b/ansible/rancher/downstream/elemental/libvirt/elemental-playbook.yml
@@ -29,21 +29,12 @@
         - hostvars['localhost'].download_url.stdout | length > 0
       retries: 3
 
-    - name: Install virt-manager
+    - name: Install libvirt packages
       ansible.builtin.apt:
-        name: virt-manager
-        state: present
-        update_cache: yes
-
-    - name: Install libvirt-daemon-system
-      ansible.builtin.apt:
-        name: libvirt-daemon-system
-        state: present
-        update_cache: yes
-
-    - name: Install libvirt-clients
-      ansible.builtin.apt:
-        name: libvirt-clients
+        name:
+          - virt-manager
+          - libvirt-daemon-system
+          - libvirt-clients
         state: present
         update_cache: yes
 
--- a/ansible/rancher/downstream/elemental/libvirt/elemental-playbook.yml
+++ b/ansible/rancher/downstream/elemental/libvirt/elemental-playbook.yml
@@ -32,9 +32,9 @@
     - name: Install libvirt packages
       ansible.builtin.apt:
         name:
-          - virt-manager
-          - libvirt-daemon-system
-          - libvirt-clients
+          - virt-manager=1:4.0.0-1
+          - libvirt-daemon-system=8.0.0-1ubuntu7.15
+          - libvirt-clients=8.0.0-1ubuntu7.15
         state: present
         update_cache: yes
 
--- a/ansible/rancher/downstream/elemental/libvirt/elemental-playbook.yml
+++ b/ansible/rancher/downstream/elemental/libvirt/elemental-playbook.yml
@@ -32,9 +32,9 @@
     - name: Install libvirt packages
       ansible.builtin.apt:
         name:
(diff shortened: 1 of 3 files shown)


ansible/rancher/downstream/elemental/libvirt/elemental-playbook.yml
```diff
@@ -35,6 +35,18 @@
...
+        name: libvirt-daemon-system
+        state: present
+        update_cache: yes
+
+    - name: Install libvirt-clients
+      ansible.builtin.apt:
+        name: libvirt-clients
+        state: present
+        update_cache: yes
+
```
floatingman: These can all be consolidated into one apt task.
```suggestion
   - name: Install libvirt packages
      ansible.builtin.apt:
        name:
          - libvirt-dev
          - libvirt-daemon-system
          - libvirt-clients
        state: present
        update_cache: yes
  

```

floatingman (commented):
Just one suggestion.

ansible/rancher/downstream/elemental/libvirt/elemental-playbook.yml
lscalabrini01: Thanks! it's better

slickwarren (dismissed):

ansible/rancher/downstream/elemental/libvirt/elemental-playbook.yml
lscalabrini01: done

floatingman (dismissed):

hamistao (dismissed):

ansible/rancher/downstream/elemental/tasks/install-elemental-tasks.yml
```diff
@@ -25,7 +25,6 @@
         create_namespace: true
         wait: true
         wait_timeout: "60s"
-        chart_version: "1.7.3"
```
hamistao, line 28: I am not sure about this, but should we keep an explicit version due to that discussion over always pinning versions?

lscalabrini01: done

floatingman, line 28: This is true. You can even make this a variable if you don't want to hard-code a version.

ansible/rancher/downstream/elemental/tasks/install-elemental-tasks.yml
```diff
@@ -39,7 +38,6 @@
...
```
floatingman, line 42: Same here. It makes sense to use a variable so you don't have to hardcode the same version multiple times.

ansible/rancher/downstream/elemental/tasks/install-elemental-tasks.yml
lscalabrini01: done!

ansible/rancher/downstream/elemental/tasks/install-elemental-tasks.yml
lscalabrini01: done!

ansible/rancher/downstream/elemental/tasks/install-elemental-tasks.yml
```diff
@@ -5,6 +5,9 @@
   vars_files:
     - vars.yaml
 
+  vars:
+    elemental_chart_version: ""
```
floatingman, line 9: Sorry to keep harping on this, but can you set a default version here? I know it will be overwritten in the Jenkins job, but for supply chain safety's sake, in case someone omits this in their Jenkins job, there's no telling which version would be installed.

floatingman (commented):
You need to set a default chart version.

ansible/rancher/downstream/elemental/tasks/install-elemental-tasks.yml
lscalabrini01: The default version is the "", it will install latest elemental chart version

floatingman: I understand, but the point of mitigating supply chain attacks is to use exact versions. But I'm alright with this if everyone else agrees.

floatingman (approved):

hamistao (approved):

ansible/rancher/downstream/elemental/tasks/install-elemental-tasks.yml
lscalabrini01: I understand your point, but during the retro we aligned on this, since it’s an Elemental chart, and we’re validating Elemental itself, it makes sense to run the tests against the latest version

floatingman: Just to note, you can override this variable in the Jenkins job.

lscalabrini01: Yes, it would be possible to override the variable in the Jenkins job if needed, adding that variable was a good catch!

PR #85, opened 2026-04-08
[UI] Add env flag to allow skip

yonasberhe23:
Adds CYPRESS_ALLOW_FILTERED_CATALOG_SKIP (default: allow skip when the chart is filtered from the catalog). Set it to false to fail the run instead of skipping in that case. 
Necessary to support https://github.com/rancher/dashboard/pull/17056

### DIFF
--- a/ansible/testing/dashboard-e2e/QUICKSTART.md
+++ b/ansible/testing/dashboard-e2e/QUICKSTART.md
@@ -85,7 +85,7 @@ export AWS_SECRET_ACCESS_KEY="..."
 
 </details>
 
-See [`vars.yaml.example`](vars.yaml.example) for all available options.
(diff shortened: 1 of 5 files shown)


izaac (approved):

izaac:
I was thinking about the approach. Let’s make sure we drive the implementation through `vars.yaml` instead of adding another environment variable, we can then use the value in the cypress config

**`vars.yaml.example`**:

`allow_filtered_catalog_skip: true`

**`templates/env.j2`**:

`CYPRESS_ALLOW_FILTERED_CATALOG_SKIP={{ allow_filtered_catalog_skip | default(true) | lower }}`

And your code will consume that. No new environment variables in the JJB.

izaac (changes requested):
Read previous comment

izaac (changes requested):
It's looking great, we just need to maintain the docs with this new var. Might be only README but I would confirm QUICKSTART to be sure.

izaac (approved):
All good, I would check the doc with `markdownlint-cli` to be safe, but this looks good.

PR #89, opened 2026-04-13
Add Play to upgrade Harvester

hamistao:
This adds a play under `ansible/harvester` to perform a Harvester upgrade from `localhost` using the Harvester API.

### DIFF
--- a/ansible/harvester/upgrade-harvester.yml
+++ b/ansible/harvester/upgrade-harvester.yml
@@ -1,0 +1,172 @@
+---
+- name: Harvester Cluster Upgrade and VM State Management
+  hosts: localhost # This playbook only uses localhost, so no inventory is needed.
+  gather_facts: false
+  vars:
+    upgrade_annotations:
+      harvesterhci.io/skip-version-check: "true"
+    rancher_releases_host: "releases.rancher.com"
+
+  tasks:
+  - name: Check all needed values are present
+    ansible.builtin.assert:
+      that:
+      - harvester_host is defined
+      - harvester_host | length > 0
+      - harvester_user is defined
+      - harvester_user | length > 0
+      - harvester_password is defined
+      - harvester_password | length > 0
+      - target_version is defined
+      - target_version | length > 0
+      - target_version is match("^v\d+\.\d+\.\d+(-(rc\d+|dev-\d{8}))?$")
+      fail_msg: "Some required variable is missing or not in the expected format"
+
+  - name: Authenticate to Harvester
+    uri:
+      url: "https://{{ harvester_host }}/v3-public/localProviders/local?action=login"
+      method: POST
+      body_format: json
+      status_code: 201
+      body:
+        username: "{{ harvester_user }}"
+        password: "{{ harvester_password }}"
+      validate_certs: "{{ skip_tls | default(false) | bool }}"
+    no_log: true
+    register: login_response
+
+  - name: Set Auth Token
+    set_fact:
+      auth_token: "Bearer {{ login_response.json.token }}"
+      token_id: "{{ login_response.json.id }}"
+    no_log: true
+
+  - name: Get running VMs before upgrade
+    uri:
+      url: "https://{{ harvester_host }}/v1/harvester/kubevirt.io.virtualmachines?exclude=metadata.managedFields"
+      method: GET
+      headers:
+        Authorization: "{{ auth_token }}"
+      validate_certs: "{{ skip_tls | default(false) | bool }}"
+    register: vm_list_resp
+
+  - name: Filter list of active VM IDs
+    set_fact:
+      running_vms: "{{ vm_list_resp.json.data | selectattr('metadata.state.name', 'equalto', 'active') | map(attribute='id') | list }}"
+
+  - name: Fetch Version Manifest from Rancher Releases
+    uri:
+      url: "https://{{ rancher_releases_host }}/harvester/{{ target_version }}/version.yaml"
+      method: GET
+      return_content: true
+    register: version_yaml_resp
+
+  - name: Create Version via Harvester API
+    uri:
+      url: "https://{{ harvester_host }}/v1/harvester/harvesterhci.io.versions"
+      method: POST
+      body_format: json
+      body: "{{ version_yaml_resp.content | from_yaml }}"
+      headers:
+        Authorization: "{{ auth_token }}"
+      validate_certs: "{{ skip_tls | default(false) | bool }}"
+      status_code: [ 201, 409 ]
+    register: version_creation
+
+  - name: Notify version existence
+    debug:
+      msg: "Version object already existed."
+    when: version_creation.status == 409
+
+  - name: Trigger Upgrade
+    uri:
+      url: "https://{{ harvester_host }}/v1/harvester/harvesterhci.io.upgrades/harvester-system"
+      method: POST
+      body_format: json
+      body:
+        type: "harvesterhci.io.upgrade"
+        metadata:
+          name: "upgrade-{{ target_version | replace('.', '-') }}"
+          namespace: "harvester-system"
+          annotations: "{{ upgrade_annotations }}"
+        spec:
+          version: "{{ target_version }}"
+      headers:
+        Authorization: "{{ auth_token }}"
+      validate_certs: "{{ skip_tls | default(false) | bool }}"
+      status_code: 201
+    register: trigger_resp
+
+  - name: Wait for upgrade to finish
+    uri:
+      url: "https://{{ harvester_host }}/apis/harvesterhci.io/v1beta1/namespaces/harvester-system/upgrades/upgrade-{{ target_version | replace('.', '-') }}"
+      method: GET
+      headers:
+        Authorization: "{{ auth_token }}"
+      validate_certs: "{{ skip_tls | default(false) | bool }}"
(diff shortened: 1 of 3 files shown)


ansible/harvester/upgrade-harvester.yml
```diff
@@ -0,0 +1,134 @@
...
+      method: POST
+      body_format: json
+      body: "{{ version_yaml_resp.content | from_yaml }}"
+      headers:
+        Authorization: "{{ auth_token }}"
+      validate_certs: false
+      status_code: [ 201, 409 ]
+    register: version_creation
+
+  - name: Notify version existence
```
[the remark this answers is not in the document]
hamistao: This runs just fine as is and my linter also does not complain.

ansible/harvester/group_vars/all.yml
floatingman, line 1: 🚨Rename this file to all.yml.example and then add all.yml to gitignore. That way, if someone puts a real password in all.yml, it won't get committed by accident if someone updates this repo.

ansible/harvester/upgrade-harvester.yml
```diff
@@ -0,0 +1,136 @@
...
+        Authorization: "{{ auth_token }}"
+      validate_certs: "{{ skip_tls | default(false) | bool }}"
+      status_code: [ 200, 502, 503, 504 ] # The node may be off during a part of the process.
+    register: upgrade_status
+    until: upgrade_status.status == 200 and (upgrade_status.json.metadata.labels['harvesterhci.io/upgradeState'] | default('')) == 'Succeeded' and (upgrade_status.json.status.conditions | selectattr('type', 'equalto', 'Completed') | map(attribute='status') | first) == 'True'
+    retries: 120
+    delay: 60
+    failed_when:
+    - upgrade_status.status == 200
+    - upgrade_status.json.status.conditions | selectattr('status', 'equalto', 'False') | list | length > 0
```
floatingman, line 114: This might cause this task to wait up to 2 hours if a cluster is truly dead. If I were you, I'd add logic to check for 5xx responses and fail after a certain amount of time (maybe 10 minutes) when clusters are unreachable.

ansible/harvester/upgrade-harvester.yml
```diff
@@ -0,0 +1,136 @@
...
+    loop: "{{ running_vms }}"
+    register: vm_start_results
+
+  - name: Wait for all VMs to return to active state
+    uri:
+      url: "https://{{ harvester_host }}/v1/harvester/kubevirt.io.virtualmachines?exclude=metadata.managedFields"
+      method: GET
+      headers:
+        Authorization: "{{ auth_token }}"
+      validate_certs: false
```
floatingman: Why is this validate_certs hardcoded to false when the ones in other tasks have conditions?

ansible/harvester/upgrade-harvester.yml
```diff
@@ -0,0 +1,136 @@
...
+  - name: Filter list of active VM IDs
+    set_fact:
+      running_vms: "{{ vm_list_resp.json.data | selectattr('metadata.state.name', 'equalto', 'active') | map(attribute='id') | list }}"
+
+  - name: Fetch Version Manifest from Rancher Releases
+    uri:
+      url: "https://{{ rancher_releases_host }}/harvester/{{ target_version }}/version.yaml"
+      method: GET
+      return_content: true
+    register: version_yaml_resp
```
floatingman, line 62: It's a good idea to validate this YAML before running the POST later. Good checks include ensuring it isn't empty and that it parses as YAML correctly.

ansible/harvester/upgrade-harvester.yml
```diff
@@ -0,0 +1,136 @@
...
+
+  - name: Start VMs that were running pre-upgrade
+    uri:
+      url: "https://{{ harvester_host }}/v1/harvester/kubevirt.io.virtualmachines/{{ item }}?action=start"
+      method: POST
+      headers:
+        Authorization: "{{ auth_token }}"
+      validate_certs: "{{ skip_tls | default(false) | bool }}"
+      status_code: 204
+    loop: "{{ running_vms }}"
```
floatingman, line 136: If one VM fails to start, it will break this loop. A better idea is to set `ignore_errors: true`, then add a task after this one that reports the status of VMs. That way, you'll know which VM failed to start.

ansible/harvester/group_vars/all.yml
```diff
@@ -0,0 +1,6 @@
+harvester_host: "2.3.4.5"
+harvester_user: "admin"
+harvester_password: "password"
+target_version: "v1.8.0-rc1"
+
+# skip_tls: false
```
floatingman: Why is this commented out?

ansible/harvester/group_vars/all.yml
```diff
@@ -0,0 +1,6 @@
+harvester_host: "2.3.4.5"
+harvester_user: "admin"
+harvester_password: "password"
+target_version: "v1.8.0-rc1"
```
floatingman: It's a good idea to validate that this is in the correct format.

floatingman (changes requested):
It'll be useful to add documentation on how to use this playbook.

ansible/harvester/upgrade-harvester.yml
hamistao: This would be risky. The update does indeed take a long time (2 hours is the same timeout used for the harvester team tests) and we know the cluster will be down for part of the process and we can't know for how long. Plus even if the update fails the node should be started again so this check would almost never be useful.

ansible/harvester/upgrade-harvester.yml
hamistao: no reason, will fix

ansible/harvester/group_vars/all.yml
hamistao: Because this is an optional variable

ansible/harvester/upgrade-harvester.yml
hamistao: The yaml syntax check is already done in a sense through `from_yaml` in the following task. We can check if it is empty but I don't see the point as the following task would also report an empty body was passed in.

ansible/harvester/upgrade-harvester.yml
floatingman: So it does return 5xx for a little while?

ansible/harvester/upgrade-harvester.yml
```diff
@@ -0,0 +1,136 @@
...
+    register: trigger_resp
+
+  - name: Wait for upgrade to finish
+    uri:
+      url: "https://{{ harvester_host }}/apis/harvesterhci.io/v1beta1/namespaces/harvester-system/upgrades/upgrade-{{ target_version | replace('.', '-') }}"
+      method: GET
+      headers:
+        Authorization: "{{ auth_token }}"
+      validate_certs: "{{ skip_tls | default(false) | bool }}"
+      status_code: [ 200, 502, 503, 504 ] # The node may be off during a part of the process.
```
floatingman, line 107: Might want to check for a 401 on the off chance that the token expires during the upgrade.

ansible/harvester/upgrade-harvester.yml
floatingman: You don't agree with this comment?

hamistao:
@floatingman As for the documentation, I don't think a README is needed here since to run this playbook one just has to fill in `all.yml` and run `ansible-playbook upgrade-harvester.yml`. So I will add comments on `all.yml.example` to make it clear which value each parameter is supposed to have. What do you think of this option?

ansible/harvester/upgrade-harvester.yml
floatingman: from_yaml doesn't do a syntax check on the YAML or check whether what's passed to it is empty, so the POST may still go through and might not return an error, or it might return a cryptic error message.

ansible/harvester/upgrade-harvester.yml
hamistao: yes, that's right

ansible/harvester/upgrade-harvester.yml
hamistao: I do, just haven't tested this yet hence why this is not marked as resolved.

ansible/harvester/upgrade-harvester.yml
hamistao: The token is created during the run with so this is not possible. Plus why would we want to not fail when getting a 401?

ansible/harvester/upgrade-harvester.yml
hamistao: To be clear, all the comments that I haven't responded are being implemented. Thanks for them btw, particularly for this one and the validate_certs one as I think they were great improvements on this playbook.

ansible/harvester/upgrade-harvester.yml
hamistao: My point is that debugging the scenario where the response comes out empty or poorly formatted is already trivial, since (I am almost sure) the body used in the request is shown by Ansible. It would be just like checking that every API call to Harvester is a properly formatted JSON, which I also don't think we need to do.

lscalabrini01 (dismissed):

floatingman (approved):

lscalabrini01 (approved):

PR #92, opened 2026-04-15
[UI] Simplify overlay branch logic

yonasberhe23:
Overlay deps now use dashboard_overlay_branch only, defaulting to master. dashboard_branch is no longer used as an implicit overlay source.


izaac (approved):
💯

PR #93, opened 2026-04-22
HA Deploy support in Go

slickwarren:


floatingman (dismissed):
There are a few suggestions.

ansible/rancher/default-ha/rancher-playbook.yml
```diff
@@ -65,20 +82,39 @@
...
       kubernetes.core.k8s:
         kubeconfig: "{{ kubeconfig_file }}"
-        src: https://github.com/cert-manager/cert-manager/releases/download/v{{ cert_manager_version }}/cert-manager.crds.yaml
+        src: /tmp/cert-manager.crds.yaml
         state: present
       retries: 5
       delay: 30
       when: cert_manager_version != ""
 
+    - name: Download cert-manager manifest
```
floatingman, line 119: I created a new role called download_verify to prevent supply chain attacks.
https://github.com/floatingman/qa-infra-automation/blob/e0f740a72ba4d97bcd4d557f3ede7c5da3af5c1f/ansible/roles/download_verify/defaults/main.yml

and here's how I use it to install cert manager
https://github.com/floatingman/qa-infra-automation/blob/e0f740a72ba4d97bcd4d557f3ede7c5da3af5c1f/ansible/rancher/default-ha/rancher-playbook.yml#L75-L117

ansible/rancher/default-ha/rancher-playbook.yml
```diff
@@ -115,6 +207,34 @@
...
+    - name: Set privateCA helm value when a CA cert is provided
+      set_fact:
+        rancher_helm_values: "{{ rancher_helm_values | combine({'privateCA': true}) }}"
+      when: ingress_tls_source_final == "secret" and tls_ca_cert_path_final != ""
+
+    - name: Merge extra helm values
+      set_fact:
+        rancher_helm_values: "{{ rancher_helm_values | combine(extra_helm_values_final, recursive=True) }}"
+      when: extra_helm_values_final | length > 0
+
```
floatingman, line 237: Why not just build all of the helm values in one task?

floatingman (commented):
Just a few suggestions.

ansible/rancher/default-ha/rancher-playbook.yml
```diff
@@ -65,20 +82,39 @@
...
+    - name: Download cert-manager manifest
+      ansible.builtin.get_url:
+        url: https://github.com/cert-manager/cert-manager/releases/download/v{{ cert_manager_version }}/cert-manager.yaml
+        dest: /tmp/cert-manager.yaml
+        mode: '0644'
+        force: true
+      retries: 5
+      delay: 30
+      when: cert_manager_version != ""
+
```
hamistao, line 128: Can't we customize the version using helm to install as well?

ansible/rancher/default-ha/rancher-playbook.yml
```diff
@@ -99,6 +135,62 @@
...
+          - tls_key_path_final != ""
+        fail_msg: "tlsSource is 'secret' but tlsCertPath or tlsKeyPath is empty"
+      when: ingress_tls_source_final == "secret"
+
+    - name: Validate Let's Encrypt inputs
+      ansible.builtin.assert:
+        that:
+          - letsencrypt_email_final != ""
+        fail_msg: "tlsSource is 'letsEncrypt' but letsEncryptEmail is empty"
+      when: ingress_tls_source_final == "letsEncrypt"
```
hamistao: Maybe we should do these first to fail early

ansible/rancher/default-ha/rancher-playbook.yml
hamistao: Good point

ansible/rancher/default-ha/rancher-playbook.yml
```diff
@@ -193,15 +309,19 @@
           username: admin
           password: "{{ bootstrap_password }}"
         method: POST
-        force_basic_auth: true
         validate_certs: false
-        status_code: 201
+        status_code: [201, 401, 403, 500, 502, 503]
```
hamistao, line 313: Why accept all of these?

ansible/rancher/default-ha/rancher-playbook.yml
```diff
@@ -193,15 +309,19 @@
...
-      retries: 5
+      retries: 18
       delay: 10
       until: rancherjson.status == 201
       when: rancher_version != ""
+
+    - name: Fail if Rancher login never succeeded
+      ansible.builtin.fail:
+        msg: "Rancher login failed after retries. Last status: {{ rancherjson.status }}, body: {{ rancherjson.json | default('N/A') }}"
+      when: rancher_version != "" and rancherjson.status != 201
```
hamistao, line 324: Instead of accepting many different status codes and then checking this, why not just limit the valid codes to 201?

ansible/rancher/default-ha/rancher-playbook.yml
slickwarren: for me, this was failing early because status_code are the list of retry conditions. the `until` block is the successful exit condition. So if these aren't added, an intermittent 500 error was preemptively failing here.

ansible/rancher/default-ha/rancher-playbook.yml
slickwarren: for me, this was failing early because status_code are the list of retry conditions. the until block is the successful exit condition. So if these aren't added, an intermittent 500 error was preemptively failing here.

ansible/rancher/default-ha/rancher-playbook.yml
slickwarren: I can add this as a follow up task if that works for you

ansible/rancher/default-ha/rancher-playbook.yml
slickwarren: I would argue its more readable like this, but if you all disagree I can consolidate.

ansible/rancher/default-ha/rancher-playbook.yml
slickwarren: I can add this as a follow up task if that works for you

floatingman (dismissed):
I'm worried that this may break when I merge my PR #82

ansible/rancher/default-ha/rancher-playbook.yml
```diff
@@ -99,6 +150,47 @@
...
+          kind: Secret
+          metadata:
+            name: tls-rancher-ingress
+            namespace: cattle-system
+          type: kubernetes.io/tls
+          data:
+            tls.crt: "{{ lookup('file', tls_cert_path_final) | b64encode }}"
+            tls.key: "{{ lookup('file', tls_key_path_final) | b64encode }}"
+      no_log: true
+      when: ingress_tls_source_final == "secret" and tls_cert_path_final != "" and tls_key_path_final != ""
```
lscalabrini01, line 177: Question: do we need to verify the tls_cert and tls_key_path here? because we alredy checking this coditional on the assert in the [line ](https://github.com/rancher/qa-infra-automation/pull/93/changes#diff-f1701f6b581251019c2c5e310ada57e5efe4ad47ca47c60b5fd28d46264ab0c3R72)

lscalabrini01 (dismissed):
Looks good, just a comment

lscalabrini01 (approved):

floatingman (approved):

ansible/rancher/default-ha/rancher-playbook.yml
hamistao: I didn't think we were expecting those 500's, if this is the case then I guess this is fine. Though probably a comment explaining this here would be nice.

ansible/rancher/default-ha/rancher-playbook.yml
hamistao: I think we never get to do any of these follow up tasks, but if it is working I am not gonna require any changes.

PR #98, opened 2026-05-14
Reconcile dashboard-e2e with PR #82 bridge architecture

izaac:
- `provision.yml`: dump `cluster_nodes_json` + run `scripts/generate_inventory.py` per workspace; derive `rancher_host` from JSON metadata.
- `install-k3s-rancher.yml`: point -i at bridge-generated inventories; drop `TF_WORKSPACE` / `TERRAFORM_NODE_SOURCE` (no longer load-bearing).
- `tfvars.j2`: emit `private_ssh_key` so the bridge can bake it into `all.vars`.
- `inventory.j2`: deleted; bridge owns inventory now.
- `README` / `QUICKSTART`: architecture + outputs updated; `cloud.terraform` noted as fallback-only for `dashboard-e2e`.

Bundled fix in `roles/download_verify`: `github_sha256sum` extractor used a substring grep against `sha256sum-amd64.txt` and pulled the first matching line, which since `v1.35.4+k3s1` is `k3s-airgap-images-amd64.tar` a wrong file. Replaced with `awk '$2 == name'` for exact, order-independent match.

Validated clean-slate on AWS: 3-way parallel tofu apply, both bridge invocations, parallel k3s installs, Rancher deploy.


ansible/roles/download_verify/tasks/main.yml
```diff
@@ -26,7 +26,7 @@
 - name: Extract expected checksum from file
   ansible.builtin.shell: |
     {% if dv_checksum_format == 'github_sha256sum' %}
-    grep "{{ dv_github_artifact_name }}" "{{ _dv_checksum_file }}" | head -1 | awk '{print $1}'
+    awk -v name="{{ dv_github_artifact_name }}" '$2 == name {print $1; exit}' "{{ _dv_checksum_file }}"
```
floatingman:
```suggestion
    grep -E "^[0-9a-f]+\s+{{ dv_github_artifact_name }}$" "{{ _dv_checksum_file }}" | head -1 | awk '{print $1}'
```

floatingman: The grep command just needs some tweaking

izaac: updated it with an small improvement to make it posix compliant, what do you think.

floatingman (approved):

izaac:
The Jenkins staging pipeline is working, and an smoke test was done locally after the `download_verify` adjustment. Merging.

PR #114, opened 2026-06-12
k3s: split-role + cp-only datastore support + rke2 harden

fmoral2:


ansible/k3s/default/k3s-playbook.yml
```diff
@@ -138,11 +138,35 @@
...
+    # Treat fqdn as absent when it equals kube_api_host (synthetic fallback at line ~35) so apiserver pick wins.
+    - name: Compute kubeconfig server URL
+      set_fact:
+        kubeconfig_server_url: >-
+          https://{{ (fqdn if (fqdn is defined and fqdn | length > 0 and fqdn != kube_api_host)
+                      else kube_api_apiserver_host) }}:6443
+      delegate_to: localhost
+      run_once: true
+      become: false
+
```
hamistao, line 163: Not sure I follow why we don't want `kube_api_host` here. But even so, perhaps instead of having to refer back to how fqdn was defined you could set `kube_api_apiserver_host` up there, before setting `fqdn` on line 35 to something like
```
kube_api_apiserver_host: "{{ lookup('env', 'FQDN') | default(fqdn) | default((groups['all'] | map('extract', hostvars)
              | selectattr('node_roles', 'defined')
              | selectattr('node_roles', 'contains', 'cp')
              | map(attribute='ansible_host') | list | first)) | default(kube_api_host, true)
```

tests/fixtures/k3s_split_role.json
```diff
@@ -0,0 +1,40 @@
...
+      "roles": ["etcd"],
+      "public_ip": "3.4.5.6",
+      "private_ip": "10.0.3.1"
+    },
+    {
+      "name": "etcd-1",
+      "roles": ["etcd"],
+      "public_ip": "3.4.5.7",
+      "private_ip": "10.0.3.2"
+    },
```
slickwarren, line 20: this will be split brain issue with 2 etcd, right? Is that intentional?

tests/fixtures/k3s_split_role.json
slickwarren, line 1: not this file but in this folder, tests/fixtures/k3s_single_master.json still declares the master as roles: ["cp"]. Under the new semantics that fixture is now an external-datastore cluster that would never come up. The test test_k3s_master_is_first_cp_only I assume will still pass because it only asserts group membership, not rendered config — so the suite gives false confidence here. Please either update that fixture/its name to reflect the new meaning, or add a comment clarifying it's an external-datastore topology.

ansible/rke2/default/rke2-playbook.yml
```diff
@@ -117,6 +117,79 @@
...
     node_token_file: "/tmp/node_token.txt"
     ansible_ssh_common_args: "-o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null"
+  pre_tasks:
+    # CIS/etcd prep; effective config merges role-specific dict + rke2_additional_config (dtf SERVER_FLAGS path).
+    - name: CIS | Resolve effective config
+      set_fact:
+        _rke2_is_server: "{{ rke2_node_role == 'master' or 'cp' in (node_roles | default([])) or 'etcd' in (node_roles | default([])) }}"
+        _rke2_srv_eff: "{{ rke2_server_config | default({}) | combine(rke2_additional_config | default({})) }}"
+        _rke2_agt_eff: "{{ rke2_agent_config | default({}) | combine(rke2_additional_config | default({})) }}"
+        rke2_is_slemicro: "{{ 'micro' in (ansible_distribution | default('') | lower) }}"
```
slickwarren: shouldn't this flag catch sles too? currently, it won't get tagged with this.

slickwarren (commented):
some feedback / questions about the test fixtures but overall lgtm

ansible/k3s/default/k3s-playbook.yml
hamistao: Or maybe even don't fallback to `kube_api_host` on line 35, use `fqdn` instead of `kubeconfig_server_url` and fallback to using `kube_api_host` instead of `fqdn` when needed.

ansible/rke2/default/rke2-playbook.yml
```diff
@@ -117,6 +117,79 @@
...
+      tags: ['cluster', 'rke2']
+
+    # CIS expression repeated since set_fact doesn't expose sibling values within the same task.
+    - name: CIS | Determine flags
+      set_fact:
+        rke2_needs_cis_sysctl: "{{ (_rke2_is_server and 'cis' in (_rke2_srv_eff.profile | default(''))) or (not _rke2_is_server and 'cis' in (_rke2_agt_eff.profile | default(''))) }}"
+        rke2_needs_etcd_user: "{{ ((_rke2_is_server and 'cis' in (_rke2_srv_eff.profile | default(''))) or (not _rke2_is_server and 'cis' in (_rke2_agt_eff.profile | default('')))) or (_rke2_is_server and (_rke2_srv_eff | dict2items | selectattr('key', 'match', '^etcd') | list | length > 0)) }}"
```
hamistao, line 137: Though I got the jist of this, we could maybe use a comment explaining this logic for readibility.

hamistao (commented):
Thanks for this! Just a few suggestions if you think they make sense.

ansible/k3s/default/k3s-playbook.yml
fmoral2: so yes, in the split role scenarios this can be the first node as etcd only for instance so we shouldnt point `kube_api_host` , and we need a " working point of contact " since this will have at config `disable-apiserver: true`

ansible/rke2/default/rke2-playbook.yml
fmoral2: sure

ansible/k3s/default/k3s-playbook.yml
hamistao: Got it, thanks for elaborating on it.

This is fine anyway then, but do you think maybe it would be simpler to set `kube_api_apiserver_host` before `fqdn` and avoid having to refer back on how `fqdn` was set?

tests/fixtures/k3s_split_role.json
fmoral2: fixed

tests/fixtures/k3s_split_role.json
fmoral2: fixed

ansible/rke2/default/rke2-playbook.yml
fmoral2: fixed

ansible/k3s/default/k3s-playbook.yml
fmoral2: done

hamistao (dismissed):

fmoral2:
qq, shouldnt we be more restrictive in the merge? like at least not allow before all comments are resolved ?

Screenshot 2026-06-17 at 11 31 36

ansible/roles/k3s_install/templates/k3s-config.yaml.j2
```diff
@@ -52,8 +66,10 @@ node-label:
...
 
-# Node type specific configuration
-{% if k3s_node_type == 'cluster-init' %}
+# Skip cluster-init when external datastore is in use (server_flags has datastore-endpoint, or master is cp-only).
+{% set _flags_blob = (server_flags | default('')) ~ "\n" ~ (role_server_flags | default('')) %}
+{% set _uses_external_datastore = ('datastore-endpoint' in _flags_blob) or (has_cp and not has_etcd) %}
+{% if k3s_node_type == 'cluster-init' and not _uses_external_datastore %}
 cluster-init: true
 {% elif k3s_node_type == 'server' and (kube_api_host | default(k3s_api_host) | default('')) != '' %}
 server: "https://{{ kube_api_host | default(k3s_api_host) }}:6443"
```
floatingman, line 77: I'm worried this kube_api_host might be set to the etcd-only node. Have you ensured that this won't happen?

docs/reference/inventory-format.md
```diff
@@ -138,8 +143,14 @@ See the [K3s BYO guide](../guides/k3s-default-byo.md#step-1-create-the-ansible-i
 
 Key requirements:
 - First node must be in the `master` group and named `master`
```
floatingman: I think this line doesn't matter anymore. The first node name no longer matters.

ansible/rke2/default/rke2-playbook.yml
```diff
@@ -117,6 +117,75 @@
...
+
+    # Try copying the shipped rke2-cis-sysctl.conf from each known path. RPM-installed RHEL/SLES use
+    # /usr/share/rke2/ or /usr/local/share/rke2/; tarball installs use /opt/rke2/share/rke2/.
+    - name: CIS | Copy rke2-cis-sysctl.conf into /etc/sysctl.d
+      copy:
+        src: "{{ item }}"
+        dest: /etc/sysctl.d/60-rke2-cis.conf
+        remote_src: true
+        mode: '0644'
+      loop:
```
floatingman, line 155: Consider adding a `break_when` clause or changing to  `first_found`; otherwise, this loop runs unconditionally, and if two paths exist, the first one is silently overwritten.

scripts/generate_inventory.py
```diff
@@ -97,6 +97,19 @@ def validate_airgap(data: dict) -> None:
         raise ValueError(f"airgap JSON missing fields: {missing}")
 
 
+def warn_if_k3s_needs_datastore(distro: str, data: dict) -> None:
```
floatingman, line 100: You didn't add a test for this function.

floatingman (commented):
Just a few comments.

ansible/roles/k3s_install/templates/k3s-config.yaml.j2
fmoral2: yes also added a new test

scripts/generate_inventory.py
fmoral2: done

ansible/rke2/default/rke2-playbook.yml
fmoral2: done I addressed  this to select the first existing file before copying,  so we copy only one path and avoid   overwriting

docs/reference/inventory-format.md
fmoral2: done

floatingman (approved):

fmoral2:
@slickwarren @hamistao  good to merge?

hamistao (approved):

hamistao:
@fmoral2 Looks good to me!

> qq, shouldnt we be more restrictive in the merge? like at least not allow before all comments are resolved ?

Personally I think this makes sense, but it would require some discussion with the team and so far we haven't planned ourselves to have this discussion. We usually wait for two approvals even though the stablished rules only require one, perhaps enforcing this convention that we already follow would be a more easily acceptable change to make.

PR #115, opened 2026-06-30
Dashboard - Fix Rancher head version sort and macOS docker detection

izaac:
Use sort -V for the head tag so the newest alpha wins instead of helm's lexical order picking a stale one. Also detect Docker Desktop's CLI and socket paths on macOS, with Linux still using /var/run/docker.sock.


ShylajaDevadiga (approved):

IsaSih (approved):
LGTM

PR #117, opened 2026-07-06
Dashboard - Add arm64 support and harden dashboard-e2e runner

izaac:
Install kubectl for the build architecture (amd64/arm64) with pinned checksums so the runner image builds on Apple Silicon. Fall back to the bundled Electron browser when Google Chrome is absent (Chrome ships amd64 only).

Harden run.sh argument parsing: run the setup stage when 'stream provision' is used so the image is built before streaming, reject unknown commands, and make destroy and build mutually exclusive with stage verbs so a mistyped or mixed invocation can no longer silently run the full pipeline or reorder stages. Unknown flags are still forwarded to ansible.

Stop the cleanup destroy loop from swallowing tofu failures: capture the real tofu exit code through the tail pipe, then surface any non-zero rc in a dedicated post-loop task after every workspace has been attempted.

Report changed honestly for tofu apply and destroy instead of hardcoding changed: true, so no-op runs no longer show phantom changes.


a-arias (approved):
LGTM

PR #119, opened 2026-07-07
fix/update documentation

floatingman:
- **Fix Rancher head version sort and macOS docker detection (#115)**
- **k3s: split-role + cp-only datastore support + rke2 harden (#114)**
- **Updated documentation to show correct prerequisites**

### DIFF
--- a/docs/prerequisites.md
+++ b/docs/prerequisites.md
@@ -7,7 +7,7 @@ Everything you need before deploying a cluster with this repository.
 | Tool | Version | Purpose | Install |
 |------|---------|---------|---------|
 | **Python** | 3.11+ | Runs Ansible and helper scripts | [python.org](https://www.python.org/downloads/) |
-| **Ansible** | core 2.16+ | Deploys K8s and Rancher onto nodes | Installed via `requirements.txt` |
+| **Ansible** | core 2.21.1 (pinned) | Deploys K8s and Rancher onto nodes | `pip install -r requirements.txt` |
 | **OpenTofu** | 1.6+ | Provisions cloud infrastructure | [opentofu.org](https://opentofu.org/docs/intro/install/) |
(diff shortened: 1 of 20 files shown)


docs/prerequisites.md
```diff
@@ -7,7 +7,7 @@ Everything you need before deploying a cluster with this repository.
 | Tool | Version | Purpose | Install |
 |------|---------|---------|---------|
 | **Python** | 3.11+ | Runs Ansible and helper scripts | [python.org](https://www.python.org/downloads/) |
```
slickwarren: ansible-core 2.21.1 requires Python ≥ 3.12

docs/prerequisites.md
```diff
@@ -33,11 +33,11 @@ ansible-galaxy collection install -r requirements.yml
...
 <summary>Alternative: using pipx</summary>
 
-If you manage Ansible with `pipx` instead of a venv, inject the dependencies into the same environment:
+If you manage Ansible with `pipx` instead of a venv, install `ansible-core` with the same pin, then inject the supporting libraries:
 
 ```bash
-pipx install ansible
-pipx inject ansible -r requirements.txt
+pipx install "ansible-core==2.21.1"
+pipx inject ansible-core pyyaml jmespath kubernetes boto3 botocore
```
slickwarren, line 40: why did this change to not use requirements.txt anymore?

slickwarren (dismissed):
overall it looks fine but some of the docs aren't quite accurate from what I can tell.

docs/prerequisites.md
floatingman: I think I just changed the ansible version without checking the python version

docs/prerequisites.md
floatingman: This has to do with the way pip resolves package names. Explicitly naming the packages fixes that clash. I also put a comment in requirements.txt

lscalabrini01 (approved):

PR #123, opened 2026-07-20
Pass systemDefaultRegistry to the Rancher Helm chart (airgap)

floatingman:
## Problem
The `airgap_rancher_helm_deploy` role defined `rancher_system_default_registry` in `defaults/main.yml` but **never passed it to the Helm chart** — the variable was a no-op. Rancher deployed with an empty `system-default-registry` setting, so `shell-image` stayed as the public `rancher/shell:<tag>` reference, which is unpullable in an airgap cluster. Any feature that runs a rancher-shell job (e.g. `kubectl.Command`, including deployment rollback in rancher/tests `WorkloadUpgradeTest`) then fails.

## Fix
Add `global.cattle.systemDefaultRegistry: "{{ rancher_system_default_registry }}"` to **both** Helm `values` blocks (signed-chart path and `rescue` fallback). The variable defaults to `""`, so non-airgap callers are unaffected.

```yaml
values:
  hostname: "{{ rancher_private_hostname }}"
  ...
  global:
    cattle:
      systemDefaultRegistry: "{{ rancher_system_default_registry }}"
```

## Verification
- Role YAML parses; `systemDefaultRegistry` referenced in both blocks.
- Live cluster: confirmed `shell-image` is **seeded at Rancher startup** from `system-default-registry` and is **not** rewritten when the setting is changed post-install — so it must be set at deploy time, which is exactly what this change does.
- Before (on a deployed airgap cluster): `system-default-registry=''`, `shell-image='rancher/shell:v0.7.1'`.

## Companion change
`rancher/tests` `Jenkinsfile.airgap-rke2-tests` sets `rancher_system_default_registry: "\${PRIVATE_REGISTRY_URL}"` in the rendered group_vars so this role variable is populated.

### DIFF
--- a/ansible/rke2/airgap/README.md
+++ b/ansible/rke2/airgap/README.md
@@ -186,6 +186,8 @@ rancher_hostname: "rancher.example.com" # Not needed if the inventory file was g
 rancher_bootstrap_password: "your-secure-password"
 rancher_image_tag: v2.12.2
 rancher_use_bundled_system_charts: true
+# Required for airgap: private registry hosting rancher/shell and system images
+rancher_system_default_registry: "privateregistry.example.com:5000"
 ```
 
 **Important Variables:**
@@ -199,6 +201,7 @@ rancher_use_bundled_system_charts: true
 | `rancher_bootstrap_password` | Initial admin password | Required |
 | `rancher_image_tag` | Rancher version to deploy | `v2.12.2` |
 | `rancher_use_bundled_system_charts` | Use bundled charts for airgap | `true` |
+| `rancher_system_default_registry` | Private registry for Rancher system images (airgap); rewrites `shell-image` to `<registry>/rancher/shell:<tag>` at install time | `""` |
 
 #### Run the Deployment
 
--- a/ansible/rke2/airgap/docs/configuration/GROUP_VARS_GUIDE.md
+++ b/ansible/rke2/airgap/docs/configuration/GROUP_VARS_GUIDE.md
@@ -139,6 +139,45 @@ rke2_agent_options: |
 
 See [RKE2 Server Config Reference](https://docs.rke2.io/reference/server_config) and [RKE2 Agent Config Reference](https://docs.rke2.io/reference/linux_agent_config) for all options.
 
+### Rancher Deployment
+
+When `deploy_rancher: true`, Rancher is installed with Helm via the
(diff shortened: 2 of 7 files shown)


floatingman:
## 🔗 Related PRs — full fix chain for the airgap `resource name may not be empty` failure

**This PR (qa-infra-automation) is the root-cause fix.**

`WorkloadTestSuite/TestDeployments/WorkloadUpgradeTest` (and any other path using `kubectl.Command`) fails in airgap because the rancher-shell helper Job can't pull its image. Three changes address different layers:

- [ ] **rancher/shepherd#598** — stop masking the real kubectl-Job failure. Error now reports `"image can't be pulled"` instead of the misleading `"resource name may not be empty"`. (Independent improvement.)
- [ ] **rancher/qa-infra-automation#123** ← this — (root cause) `airgap_rancher_helm_deploy` defined `rancher_system_default_registry` but never passed it to the chart; now sets `global.cattle.systemDefaultRegistry` at install time so `shell-image` resolves to the private registry.
- [ ] **rancher/tests#771** — supplies the value: renders `rancher_system_default_registry: "${PRIVATE_REGISTRY_URL}"` in the airgap-rke2-tests group_vars.

**Dependency:** tests#771 sets the variable; this PR makes the role consume it. Setting the variable alone (tests#771) is a harmless no-op until this PR lands.

**Also required (ops):** mirror `rancher/shell:<tag>` to `<registry>/rancher/shell:<tag>` — the containerd `proxycache`/`quaycache` mirrors do not cover this verbatim reference.

floatingman:
**Update:** added documentation and template coverage for `rancher_system_default_registry` (commit 2):

- **New:** `roles/airgap_rancher_helm_deploy/README.md` — the deploy playbook README referenced "that role's README" which didn't exist. Documents all role variables, with `rancher_system_default_registry` called out for airgap (shell-image rewrite at install time, image-mirror prerequisite, and why post-install patching is insufficient).
- `inventory/group_vars/all.yml.template` — add the variable so copied inventories set it (the live `all.yml` example already had it).
- `rke2/airgap/README.md` — add it to the "Deploy Rancher" example snippet and the Important Variables table.
- `rke2/airgap/playbooks/deploy/README.md` — add it to the Variables table and link the new role README.
- `roles/airgap_rancher_helm_deploy/defaults/main.yml` — clarify the comment (value is passed to the chart as `global.cattle.systemDefaultRegistry`).

All YAML validated. No behavior change — docs/templates only (the code fix is in commit 1).

lscalabrini01 (approved):

PR #124, opened 2026-07-21
Add remote S3-backend state cleanup (infra-*-remote targets)

floatingman:
## Problem

`make infra-ls`, `infra-scan`, and `infra-nuke` scan **only local** `terraform.tfstate` files:

```makefile
done < <(find tofu -name "terraform.tfstate" 2>/dev/null | sort);
```

But the modules use an **S3 backend** (the repo default), so local state files don't exist — those targets silently find nothing. The real state lives in S3 under `env:/<workspace>/<key>`, which means there is no way to list, destroy, or clean up S3-backed infrastructure through the existing tooling. (Discovering stale/orphaned workspace state — common when CI destroys infra but leaves the state file behind — is also unsupported.)

## Solution

Add `tofu/scripts/remote-state.sh`, which reads state directly from S3 (backend bucket/key/region are auto-discovered from `<module>/backend.tf`), and four Make targets that wrap it.

| Make target | Subcommand | Purpose |
|---|---|---|
| `infra-ls-remote` | `list` | List S3-backed workspaces that have managed resources |
| `infra-nuke-remote` | `destroy` | Destroy all S3-backed workspaces for the current module |
| `infra-empty-folders` | `empty-folders` | Remove workspace folders whose state has **0** resources |
| `infra-stale-folders` | `stale-folders` | Remove state whose AWS resources are **verified gone** |

### `stale-folders` (the interesting one)

A workspace's state can still *list* resources that no longer exist in AWS — e.g. a CI job that destroyed the infra but left the state file, or a `tofu destroy` that errored due to a region mismatch. `stale-folders` handles this safely:

- For each workspace it reads the **real region from the state itself** (load-balancer ARN / instance AZ), so it is unaffected by the `aws_region` in `terraform.tfvars`.
- It runs `aws ec2 describe-instances` and flags a workspace **stale only if all its instances are confirmed gone**.
- It never deletes a workspace that still has live instances — safe by construction.
- This avoids fighting `tofu destroy` (which errors on stale/wrong-region state) when the infra is already gone.

### Design choices

- **No bucket deletion.** Per discussion, buckets are never deleted (especially in shared accounts). `infra-nuke-remote` destroys workspaces and stops; `infra-empty-folders` / `infra-stale-folders` remove state *objects*, leaving the bucket intact.
- Read-only by default; mutations require explicit `DELETE=yes` (and a typed confirmation, or `AUTO_APPROVE=yes`).
- Scope with `NUKE_FILTER=<regex>` on workspace name.

## Usage

```bash
make infra-ls-remote                              # list S3-backed workspaces with resources
make infra-nuke-remote                            # destroy all S3-backed workspaces
make infra-empty-folders NUKE_FILTER='dnew'       # list your 0-resource folders (read-only)
make infra-empty-folders NUKE_FILTER='dnew' DELETE=yes
make infra-stale-folders                          # which workspaces are gone in AWS? (read-only)
make infra-stale-folders DELETE=yes               # remove confirmed-stale states
```

## Docs

- `docs/reference/makefile.md` — new table rows + a **"Local vs. remote state cleanup"** section
- `docs/reference/troubleshooting.md` — entries for "`infra-nuke` finds nothing" and "`tofu destroy` errors (stale state)"
- `tofu/scripts/README.md` — full `remote-state.sh` reference

## Testing

- `bash -n tofu/scripts/remote-state.sh` passes; all four Make targets parse.
- Validated live (read-only) against the shared state bucket: `infra-ls-remote` correctly enumerates S3-backed workspaces that `infra-ls` cannot see; `infra-stale-folders` correctly verified 38 CI workspaces whose instances/NLBs/route53 records are all gone in AWS before flagging them stale.

## Files

- `tofu/scripts/remote-state.sh` (new)
- `Makefile`
- `docs/reference/makefile.md`
- `docs/reference/troubleshooting.md`
- `tofu/scripts/README.md`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -116,6 +116,10 @@ help: ## Show this help message
 	@echo "  infra-ls            List ALL active infrastructure across every module/workspace"
 	@echo "  infra-scan          Detailed scan of all infrastructure with resource counts"
 	@echo "  infra-nuke          Destroy ALL active infrastructure (end-of-day cleanup)"
+	@echo "  infra-ls-remote     List remote (S3-backed) workspaces with resources (current module)"
+	@echo "  infra-nuke-remote   Destroy all remote (S3-backed) workspaces for the current module"
+	@echo "  infra-empty-folders Remove empty workspace folders (0-resource states) from the current module's bucket"
+	@echo "  infra-stale-folders  Remove STALE workspace folders (resources gone in AWS) — verified against AWS"
 	@echo ""
 	@echo "CLUSTER (Ansible):"
 	@echo "  cluster             Install Kubernetes cluster"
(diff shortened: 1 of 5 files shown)


floatingman:
Superseded by #126 — consolidated this with the other airgap work into a single PR to reduce review load. All changes here are included in #126.

PR #136, opened 2026-08-02
[Dashboard] Adapt dashboard-e2e to Cypress 15 and add a --spec flag

izaac:
## Summary

Upstream `rancher/dashboard` moved master from Cypress `11.1.0` to `15.19.0` and `@cypress/grep` from 3.x/4.x to `6.0.2`. The v6 exports map dropped the `src/*` entrypoints this playbook and the release branches import, so the runner broke.

This adapts to the new layout without dropping support for older branches, adds a `--spec` flag, and fixes four defects found along the way.

| Fix | Impact |
|---|---|
| Grep settings move to `Cypress.expose` on v5+ | **Tag filtering was completely inert under Cypress 15.** Every test in a selected spec ran regardless of its tags |
| `@cypress/grep` debug output is now opt in | **AWS keys, the Azure client secret and an SSH private key were printed into every console log** |
| Docker socket gid is read from inside a container | Every macOS Docker Desktop run failed at the image build step |
| Reporter chain degrades when master-only deps are absent | `--dashboard-dir` on a release branch could not load a reporter |

## Required actions on the Jenkins side

1. Bump `cypress_version` to `15.19.0` in every job's `VARS_YAML_CONFIG`. There is no playbook default and setup fails on an undefined variable.
2. Expect test counts to drop. Jobs currently run every test in each selected spec file; after this a `@adminUser` job runs only `@adminUser` tests.

Otherwise Jenkins is unaffected. `cypress/jenkins/init.sh` never calls `run.sh`, passes no arguments to the Cypress container and never sets `HOST_UID` (verified: zero occurrences of `HOST_UID`, `HOST_GID`, `--user`, `userns`). The reporter change is a no-op there because the clone path always overlays master's dependency manifests.

## The tag filtering defect

The most important finding, and the reason this PR is larger than the original bump.

`@cypress/grep` v3 and v4 read their settings from `config.env`. **v6 reads them from `config.expose`**, the Cypress 15 replacement, on both the plugin side (`expose.grepTags`) and the browser side (`Cypress.expose('grepTags')`). Tags left in `env` are never seen.

- Symptom: the plugin logs `Cypress config expose object: {}` and never prints `filtering using tag(s)`. No test level filtering happens at all.
- Consequence: every test inside a selected spec file ran, whatever its tags. A `@adminUser` job also ran the `@standardUser` tests in the same file. The spec level pre-selection done by `grep-filter.ts` still worked, which is what hid the problem.
- Proof: `login.spec.ts` is tagged `@generic, @adminUser, @standardUser`. Running it with `cypress_tags: "@userMenu"` executed all 3 tests. It should execute none.
- Fix: attach the grep settings to `expose` when the installed grep exposes `@cypress/grep/plugin` (v5+), and to `env` otherwise. Cypress 11 rejects an unknown top level `expose` key, so the choice cannot be unconditional.

Upstream `cypress/base-config.ts` on master still places `grepFilterSpecs` and `grepOmitFiltered` in `env`, so local `yarn cy:run` has the same gap. 

## Validation

49 recorded runs across five environments. Every code path in `run.sh` and `init.sh` was exercised against real AWS infrastructure.

| Area | Result |
|---|---|
| Smoke matrix, release-2.13 / 2.14 / 2.15 | Provision, clone+tag, clone+spec, clone+filter, dir+spec, destroy. All green |
| Jenkins path, `init.sh` on master | 17/17 tests, exit 0, playbook `ok=74 failed=0` |
| `job_type: existing` | Pass |
| `qase_enabled: true` | Pass, live API, run created and closed |
| macOS arm64 end to end | Pass after the socket gid fix |
| Static analysis, unit and isolated checks | `shellcheck` clean, `ansible-lint` at the `origin/main` baseline, 12/12 grep-compat fixtures |

The `clone+filter` stage is the regression guard: the same spec with a tag it does not carry must select 0 tests. Before the `expose` fix it ran all 3.

<details>
<summary>Full smoke matrix, release-2.13 / 2.14 / 2.15</summary>

### Final smoke matrix

Real AWS infrastructure, one Rancher per release branch, destroyed between branches. Each branch runs four Cypress stages plus provision and destroy.

| Stage | What it proves |
|---|---|
| `clone+tag` | Cypress 15 + grep 6.0.2 against release branch specs, tag pre-filter selects spec files |
| `clone+spec` | `--spec` is honoured end to end and the named spec's tests execute |
| `clone+filter` | same spec with a tag it does not carry, expects 0 tests, proves grep still filters at test level |
| `dir+spec` | `--dashboard-dir` with the branch's own Cypress 11 and grep 3.x/4.x |

| Branch | Rancher | provision | clone+tag | clone+spec | clone+filter | dir+spec | destroy |
|---|---|---|---|---|---|---|---|
| release-2.13 | v2.13-head | Pass | 19 pass / 1 fail | 2 pass / 1 fail | not run | rejected by preflight, as designed | Pass |
| release-2.14 | v2.14-head | Pass | 20 tests, 19 pass / 1 fail | 3 tests, 2 pass / 1 fail | **0 tests, rc=0** | 3 tests, 2 pass / 1 fail | Pass |
| release-2.15 | v2.15-head | Pass | 20 tests, 19 pass / 1 fail | 3 tests, 0 pass / 1 fail | **0 tests, rc=0** | 3 tests, 0 pass / 1 fail | Pass |

`clone+filter` returning 0 tests is the pass condition. Before the `expose` fix the same run executed all 3 tests in the file, which is how the dead filter was found.

The log lines that confirm each fix, taken from the release-2.14 run:

```
Grep: passing filter settings through 'expose' (tags: @userMenu+-@prime+-@noVai)
@cypress/grep: filtering using tag(s) "@userMenu+-@prime+-@noVai"
@cypress/grep: non-matching tests will be omitted from results (not skipped)
Tests: 0
Reporters: no tests ran, skipping report generation.
```

```
Reporters: cypress-multi-reporters or mocha-junit-reporter is not installed.
           Falling back to cypress-mochawesome-reporter. No JUnit XML will be produced.
Grep: passing filter settings through 'env' (tags: @adminUser+-@prime+-@noVai)
Cypress: 11.1.0
```

All remaining failures are branch drift, not pipeline defects. Each is an assertion or a Cypress API error inside the branch's own spec, never a module resolution or config error:

| Failure | Branches | Cause |
|---|---|---|
| `account-api-keys.spec.ts` suite-level override rejected | 2.13, 2.14, 2.15 | uses `testIsolation: 'off'`, a value Cypress removed in v12 |
| login lands on `#cluster-events` instead of `/dashboard/home` | 2.13, 2.14, 2.15 | dashboard behaviour differs from the assertion on these branches |
| `cy.task('getHostStats')` not registered | 2.15 | task exists only in the branch's own plugin set |

After the matrix: 0 EC2 instances, 0 Route53 records, all three tofu workspaces at 0 resources, `vars.yaml` restored.

</details>

<details>
<summary>Jenkins path simulation on master</summary>

### Jenkins path simulation

`cypress/jenkins/init.sh` is the real entry point on Jenkins and never calls `run.sh`, so it was run directly with the environment the Jenkinsfile provides. Only one deviation from production: the docker socket mount was parameterised so it could point at a rootless podman socket.

| Setting | Value |
|---|---|
| `QA_INFRA_BRANCH` | this branch, cloned by `init.sh` itself |
| `dashboard_branch` | `master` |
| `rancher_image_tag` | `head` via `rancher-com-alpha` |
| `cypress_tags` | `@standardUser+@userMenu` |
| `create_initial_clusters` | `false` |

Result: **17 tests, 17 passing, Cypress exit 0**, playbook `ok=74 failed=0`.

What it proves, none of which the `run.sh` matrix covered:

- The `master` branch, which is the default Jenkins target.
- The buffered path. `init.sh` runs the playbook by tag and then the Cypress container directly, with no `--user`, no `HOME` override and no `HOST_UID`.
- The `@standardUser` branch. `Update TEST_USERNAME for @standardUser` fired, `rancher_user_setup` created and verified `standard_user`, and Cypress logged in as it.
- `create_initial_clusters: false`.
- The runner image built from `Dockerfile.quickstart` rather than by `run.sh`.

The lines that matter:

```
cypress_tags=@standardUser+@userMenu+-@prime+-@noVai
grep-filter: matched 3 spec(s)
Username: standard_user
Grep: passing filter settings through 'expose' (tags: @standardUser+@userMenu+-@prime+-@noVai)
@cypress/grep: filtering using tag(s) "@standardUser+@userMenu+-@prime+-@noVai"
account-api-keys.spec.ts   83ms   1   1
All specs passed!  00:45  17  17
```

`account-api-keys.spec.ts` running exactly 1 test is the point. Before the `expose` fix the same run executed every test in that file.

</details>

<details>
<summary>job_type=existing, Qase and macOS arm64</summary>

### Remaining configuration paths

Closed against a single provisioned Rancher to keep the cost down. One `job_type: recurring` provision, then two runs against it by hostname.

| Path | Result |
|---|---|
| `job_type: existing` | Pass. Provisioning skipped entirely, `rancher_host` taken from `vars.yaml`, 2/2 tests |
| `qase_enabled: true` | Pass. Live Qase API, project SANDBOX. Run 114 created, both results posted, run closed |
| macOS arm64 end to end | Pass after the socket gid fix. Cypress 15.19.0, electron fallback, 2/2 tests, playbook `ok=52 failed=0` |

Qase availability depends on the mode, not the branch. `cypress-qase-reporter` is declared on master and release-2.15 only. Clone runs overlay master's manifests onto every release branch, so Qase works on all of them, and Jenkins is always a clone run. A `--dashboard-dir` run on release-2.14 keeps the branch manifest, has no Qase reporter, and now logs that reporting is unavailable instead of throwing.

The macOS run is what surfaced the socket gid defect. It is a real regression for every Docker Desktop user, not an artifact of the test setup.

</details>

<details>
<summary>Static analysis</summary>

### Static analysis

| Check | Result |
|---|---|
| `shellcheck -s sh run.sh` | Clean |
| `shellcheck -s bash files/cypress.sh` | SC1091 only (sources `utils.sh`) |
| `ansible-playbook --syntax-check` | rc=0 |
| `ansible-lint` | 4 findings, **identical to the `origin/main` baseline**. No new lint debt |
| `run.sh` guard matrix | 14/14 (arg parsing, conflicting verbs, `--spec` rejection cases) |
| Secrets scan of the push payload | Clean |
| README vs code audit | 8 claims checked, 7 verified, 1 partial (since corrected) |

</details>

<details>
<summary>Unit and isolated checks</summary>

### Unit / isolated

| Check | Result |
|---|---|
| `grep-filter.ts` utils resolution vs real npm installs of 6.0.2 / 4.1.1 / 3.1.5 | 3/3, v6 resolves `dist/utils.js`, v3/v4 resolve `src/utils.js` |
| `cypress.config.jenkins.ts` plugin resolution, same three versions | 3/3, v6 uses `/plugin`, v3/v4 fall back to `src/plugin` |
| Narrowed catch: injected `TypeError` must surface | Surfaces as `TypeError`, not masked |
| grep-compat rewrite vs real `e2e.ts` from master / 2.13 / 2.14 / 2.15 | master no-op, all three release branches rewritten, idempotent |
| `cypress.sh` arg dispatch | 10/10, including `--browser` no longer disabling the tag pre-filter |
| Tag normalization vs real `@cypress/grep` 6.0.2 | 8 input/repo pairs; empty now yields `-@prime+-@noVai` (matches) instead of `+-@prime+-@noVai` (matches nothing) |
| `ln` behaviour matrix | 3/3; a developer's real `node_modules` survives intact |

</details>

<details>
<summary>Live on Linux with rootless Podman</summary>

### Live, Linux + rootless Podman

Against a provisioned Rancher 2.15.0-alpha22 on AWS.

| Scenario | Result |
|---|---|
| V1: `create_initial_clusters=true` + remote clone | ok=63, failed=0 |
| V2: `create_initial_clusters=true` + `--dashboard-dir` | Pass |
| V3: `create_initial_clusters=false` + remote clone | ok=59, failed=0 |
| V4: `create_initial_clusters=false` + `--dashboard-dir` | ok=55, failed=0 |
| Jenkins-conditions sim (root container, no `--user`/`HOME`/`HOST_UID`) | Pass on release-2.14 + master overlay |
| Jenkins cypress container, no arguments | grep-filter matched 121 specs for `@adminUser+-@prime+-@noVai`, 0 prime specs leaked |
| `stream --spec` single spec | 2/2 tests passed, exit 0 |
| `--spec` glob | 3 specs, 20/20 passed |
| `--spec` comma list | 2 specs, 19/19 passed |
| `run.sh` verbs: `results` / `clean` / `build` / destroy-guard | Pass |
| Full provision → destroy on real AWS | Destroy released all 3 tofu workspaces; 0 resources, 0 EC2 instances left |

</details>

<details>
<summary>Live on macOS arm64 with Docker</summary>

### Live, macOS arm64 + Docker 29.6.1

| Scenario | Result |
|---|---|
| Guard matrix under BSD `/bin/sh` | 10/10 |
| Flag wiring trace | `HOST_UID`/`HOST_GID` populated, no `keep-id`, socket gid granted, async dir wired |
| `Dockerfile.ci` arm64 branch + pinned arm64 kubectl checksum | Pass (first time this path has been exercised) |
| arm64 Chrome absent → electron fallback | Pass |
| Legacy `@cypress/grep` v3/v4 resolution inside a real built image | Pass |
| `Dockerfile.quickstart` runner image on arm64 | Pass |
| `stream --spec` end-to-end | `--spec` reached the container verbatim |
| Buffered `run.sh test` (the path that broke on Podman) | Ran `run-tests.yml` fully with `USER_ARG` active |
| Artifact ownership after a real Cypress run | 0 root-owned files in the checkout |
| Non-interactive ssh run after the stdin fix | Completes; previously died before starting a container |

</details>

<details>
<summary>Changes in detail</summary>

#### Cypress 15 / @cypress/grep v6

- Resolve the grep plugin and utils from either the v5+ (`dist/`, exports map) or the legacy v3/v4 (`src/`) layout. The fallback only triggers on a genuine `MODULE_NOT_FOUND` / `ERR_PACKAGE_PATH_NOT_EXPORTED`, so a real plugin fault surfaces as itself instead of being reported as a module-resolution failure.
- Rewrite the legacy `@cypress/grep/src/support` import during setup when the overlaid grep major is 5 or newer, so release branch specs keep working against master's libraries. Skipped for `--dashboard-dir`, where deps and specs are self-consistent.
- Bump the pinned `cypress_version` to `15.19.0`.

#### Reporter chain degrades instead of failing

`cypress.config.jenkins.ts` names `cypress-multi-reporters` and `mocha-junit-reporter`, which Cypress resolves by name at runtime. Those packages exist only in master's dependency set. Clone runs overlay master's manifests onto release branches and always have them; a `--dashboard-dir` run keeps the local checkout's manifest and does not.

- Before: `Error loading the reporter: cypress-multi-reporters` on every `--dashboard-dir` run against release-2.14 or release-2.15.
- After: falls back to `cypress-mochawesome-reporter`, which every branch ships, and logs that no JUnit XML will be produced. Qase is disabled the same way when its reporter is absent, rather than throwing from the `after:run` hook.
- Jenkins is unaffected: it always takes the clone path, which overlays master's deps.

#### Credentials no longer printed into test logs

`cypress.sh` set `DEBUG=@cypress/grep` unconditionally. The plugin's debug channel dumps the entire Cypress env object, which carries `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, `AZURE_CLIENT_SECRET` and the base64 encoded SSH private key for the custom node. Every archived Jenkins console log contained them in clear text.

Debug output is now opt in through `CYPRESS_GREP_DEBUG=true`, and prints a warning when enabled. The useful `filtering using tag(s)` line is a plain `console.log` in the plugin and is still shown.

#### New `--spec` flag

- `./run.sh stream --spec <path>` runs a single spec, a glob, or a comma-separated list.
- Validated on the host when the dashboard directory already exists, and again inside the container. Globs and comma lists defer to Cypress.
- Requires `stream`, because the buffered test stage runs the container without extra arguments and would silently run the full suite.
- Only an explicit `--spec` replaces the tag pre-filter. Other forwarded flags such as `--browser` no longer disable it.

#### Runner fixes found while testing

| Fix | Symptom before |
|---|---|
| Pass grep settings through `expose` on grep v5+ | Tag filtering did nothing. Every test in a selected spec ran regardless of its tags |
| Make the grep support rewrite fail loudly | Four inputs made the rewrite exit 0 without doing anything, leaving filtering dead: a version spec with no dot, a manifest with no grep entry, registration living outside `cypress/support/e2e.ts`, and a file that binds `register` but never calls it |
| Read the docker socket gid from inside a container | On macOS Docker Desktop every run died at the image build step with `permission denied ... /var/run/docker.sock`. The host path is a symlink owned `root:daemon`, and Docker Desktop presents the socket as `root:root` inside the container whatever the host says, so no host derived gid can be correct |
| Fall back to `cypress-mochawesome-reporter` when the multi-reporter chain is absent | `--dashboard-dir` on any release branch died with `Cannot find module 'cypress-multi-reporters'` |
| Make `@cypress/grep` debug output opt in | AWS keys, the Azure client secret and the SSH private key were printed into every console log |
| Skip report generation when no test ran, and catch report errors | `cypress-mochawesome-reporter` threw `RangeError: Invalid array length` on an empty selection, failing an otherwise clean run |
| Reject a `--dashboard-dir` checkout without `cypress/package.json` | release-2.13 keeps test deps in the root manifest, so the run died 53 tasks in at the overlay verification step |
| Check the runner image carries the requested Cypress binary | `run.sh test` and `run.sh stream test` skip the image build, so changing `cypress_version` left the npm package and the binary on different versions and Cypress reported a missing binary |
| Bind container stdin only when a terminal exists | Any non-interactive caller (ssh batch, cron) died with `cannot attach stdin to a TTY-enabled container` |
| Set `ANSIBLE_ASYNC_DIR` explicitly | With `HOME=/tmp` under Podman `keep-id`, every provision failed in ~7s with `could not find job` |
| Export `HOST_UID`/`HOST_GID` only on Docker | Rootless Podman remapped a bare `--user` to a subuid that could not write the mounted dashboard; `cypress.sh` died at `ln -sf ... Permission denied` |
| Normalize adjusted Cypress tags | Empty `cypress_tags` produced `+-@prime+-@noVai`, which matches no test at all |
| Leave a real `node_modules` directory alone | `ln -sf` nested a dangling link inside it on repeat `--dashboard-dir` runs |

#### Known limitation: `--dashboard-dir` needs release-2.14 or newer

The dependency overlay is deliberately skipped for local checkouts, because the playbook mounts the operator's working tree and must not rewrite its manifests. release-2.13 keeps the test dependencies in the root `package.json` and has no `cypress/package.json`, so it cannot satisfy `cypress.config.jenkins.ts`. `run.sh` now rejects such a checkout with exit 2 and points at the clone path instead. Documented in `README.md`.

#### Known, out of scope

Re-provisioning against an already-bootstrapped Rancher fails with HTTP 401 in `rancher_auth`: the role authenticates with the bootstrap password and then rotates the admin password away from it, destroying its own precondition. That role is untouched by this branch and is shared by five pipelines, so it gets its own [issue](https://github.com/rancher/qa-infra-automation/issues/137).

</details>

## Before merging

- [x] Smoke matrix green on release-2.13, release-2.14 and release-2.15
- [x] Jenkins path simulation green on master
- [x] `job_type: existing`, `qase_enabled: true` and macOS arm64 all green
- [x] Zero EC2 instances and all tofu workspaces at 0 resources afterwards
- [ ] Jenkins job owners bump `cypress_version` to `15.19.0` in `VARS_YAML_CONFIG`

<details>
<summary>Operational notes for reviewers</summary>

### Operational notes for reviewers

| Topic | Detail |
|---|---|
| Tofu workspaces | `rancher-server`, `importcluster` and `customnode` are shared names, so only one Rancher can exist at a time. Destroy before the next provision. |
| Stale outputs | `outputs/rancher-vars.yaml` is read by the "Recover rancher_host from tofu state" pre-task. Clear `outputs/` between branches or a stale prefix and fqdn are reused. |
| Security group | `ui-qa-team-sg` needs both `22` and `6443` from the operator IP, because the playbook makes controller side `kubectl` and `helm` calls to `<public-ip>:6443`. |
| Runner image | `dashboard-test:0` bakes one Cypress binary, chosen by `cypress_version` at build time. `run.sh test` and `run.sh stream test` skip the build. |
| Log parsing | Cypress wraps its summary counts in ANSI SGR escapes, so any parser must strip `\x1b\[[0-9;]*m` before matching `Passing:` and `Failing:`. |

</details>


yonasberhe23 (approved):
LGTM!

MSpencer87 (approved):

PR #158, opened 2026-08-06
[cluster_nodes] Add optional dedicated SSH SG to avoid prefix-list propagation lag

floatingman:
## What

Adds an opt-in **dedicated SSH security group** to the `cluster_nodes` module, attached alongside the existing `aws_security_group`, so SSH (22) from a stable CIDR (jumpbox/bastion) is authorized via a **plain CIDR rule** instead of only a managed prefix list.

## Why

When the jumpbox/bastion IP is authorized for SSH only through a managed prefix list, AWS propagates prefix-list SG rules to each new instance's ENI **asynchronously**. On freshly-launched instances one random ENI can lag or land on a stale list version and silently drop SSH SYN packets to that node — surfacing as _"one node randomly unreachable on SSH right after `tofu apply`"_: host firewall open, VPC-internal SSH working, TCP connect timeout from the jumpbox, with a **different node affected each build**.

Plain CIDR SG rules realize on the ENI immediately at launch, eliminating this flaky failure mode.

AWS SGs are additive (a flow is allowed if any attached SG allows it), so this layers on top of `aws_security_group` without touching the existing RKE2/Rancher port matrix. The SG is module-owned and destroyed on `tofu destroy`.

## Changes

- `variables.tf`: `create_ssh_security_group` (bool, default `false`), `ssh_allowed_cidrs` (list(string))
- `main.tf`: `aws_security_group.ssh`, `data.aws_vpc.selected` (auto-allows VPC-internal SSH), instance `vpc_security_group_ids` wiring
- `outputs.tf`: `ssh_security_group_id`
- `README.md`: documents the feature, the symptom it fixes, and when to enable it

## Usage

```hcl
create_ssh_security_group = true
ssh_allowed_cidrs         = ["203.0.113.10/32"]   # jumpbox / bastion / office egress
```

Opt-in; existing users see no change (`create_ssh_security_group` defaults to `false`). VPC-internal SSH is allowed automatically regardless of this setting.

## Validation

- `tofu validate` → `Success! The configuration is valid.`
- Verified end-to-end on a live cluster: with the feature enabled, the stable `/32` is realized on every ENI at launch and the random-node SSH timeout no longer reproduces.

### DIFF
--- a/tofu/aws/modules/cluster_nodes/main.tf
+++ b/tofu/aws/modules/cluster_nodes/main.tf
@@ -50,12 +50,64 @@ resource "aws_key_pair" "ssh_public_key" {
   public_key = file(var.public_ssh_key)
 }
 
+# Dedicated SSH security group with stable CIDR rules.
+#
+# Why this exists: when SSH access for the jumpbox/bastion is granted ONLY via a
+# managed prefix list (e.g. a churning "QA public IPs" list), AWS propagates the
+# prefix-list entries to each new instance's ENI asynchronously. On freshly
+# launched instances one ENI can lag or land on a stale list version and silently
+# drop the jumpbox's SSH SYN to that node until propagation catches up - so one
+# random node per build looks "unreachable" for minutes. Plain CIDR SG rules, by
+# contrast, are realized on the ENI immediately at launch, so SSH from the
+# jumpbox always works.
+#
+# This SG is attached ALONGSIDE var.aws_security_group (AWS SGs are additive: a
+# flow is allowed if ANY attached SG allows it), so the existing RKE2/Rancher
+# port matrix in the shared SG is left untouched.
+resource "aws_security_group" "ssh" {
+  count       = var.create_ssh_security_group ? 1 : 0
+  name        = "tf-${var.aws_hostname_prefix}-ssh"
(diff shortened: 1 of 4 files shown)


lscalabrini01 (approved):

tofu/aws/modules/cluster_nodes/main.tf
```diff
@@ -50,12 +50,64 @@ resource "aws_key_pair" "ssh_public_key" {
...
+# prefix-list entries to each new instance's ENI asynchronously. On freshly
+# launched instances one ENI can lag or land on a stale list version and silently
+# drop the jumpbox's SSH SYN to that node until propagation catches up - so one
+# random node per build looks "unreachable" for minutes. Plain CIDR SG rules, by
+# contrast, are realized on the ENI immediately at launch, so SSH from the
+# jumpbox always works.
+#
+# This SG is attached ALONGSIDE var.aws_security_group (AWS SGs are additive: a
+# flow is allowed if ANY attached SG allows it), so the existing RKE2/Rancher
+# port matrix in the shared SG is left untouched.
```
slickwarren, line 66: this seems like overexplaining / could be a shorter explaination

slickwarren (approved):
one nonblocking comment

PR #165, opened 2026-08-07
feat(airgap): self-provision security group when none supplied

floatingman:
## Summary

All security groups were removed in AWS, which broke the airgap OpenTofu module: it previously only **consumed** a pre-existing security group via `var.aws_security_group` (shared by the bastion, registry, and every airgap node), so `tofu apply` failed on the dangling SG reference.

This makes the module **self-contained** — it now provisions its own security group inside `var.aws_vpc` when none is supplied, while remaining backward-compatible for callers that pass an existing group.

## Changes

- **`main.tf`** — added `aws_security_group.airgap` (created in the airgap VPC, gated by `count`) and a `locals` block that selects it when `var.aws_security_group` is empty; rewired the bastion, registry, and `airgap_nodes` modules to `local.security_group_ids`; removed the now-duplicate standalone `ports` local.
- **`variables.tf`** — `aws_security_group` is now optional (`default = []`); added `allowed_ssh_cidr` and `allowed_lb_cidr` (both default `["0.0.0.0/0"]`) so source ranges are lock-down-able.
- **`outputs.tf`** — exposed `security_group_id` (null when a caller supplies their own group).
- **`README.md`** — documented the auto-provisioning and made the sample `aws_security_group` optional.

## Rules on the new SG

| Direction | Protocol / Port | Source | Purpose |
|---|---|---|---|
| ingress | TCP 22 | `allowed_ssh_cidr` | Operator SSH to the bastion |
| ingress | all (`-1`) | `self` | bastion↔nodes + node↔node RKE2/Rancher comms (etcd, kube-api, flannel VXLAN, …) |
| ingress | TCP 80/443/6443/9345 | `allowed_lb_cidr` | NLB listener ports (NLBs carry no SG, so the targets must allow these directly) |
| egress | all (`-1`) | `0.0.0.0/0` | Outbound (bastion image/tarball pulls, etc.) |

Airgap isolation is **preserved**: it is still enforced at the subnet level (the airgap subnet has no IGW route), so the permissive egress does not break airgap-ness for nodes in that subnet. This is noted in an inline comment.

## Validation

- `tofu validate` → **Success**
- `tofu plan -refresh=false -target=aws_security_group.airgap` → `Plan: 1 to add` — creates `aws_security_group.airgap` (named `<prefix>-airgap`) in the airgap VPC with all seven rules above.

### DIFF
--- a/tofu/aws/modules/airgap/README.md
+++ b/tofu/aws/modules/airgap/README.md
@@ -145,7 +145,11 @@ aws_access_key        = "key"
 aws_secret_key        = "secretkey"
 aws_ami               = "ami-"
 instance_type         = "t3.xlarge"
-aws_security_group    = ["sg-"]
+# aws_security_group is optional. When omitted (or set to []) the module
+# provisions its own security group inside var.aws_vpc that opens SSH, the
+# intra-group traffic the RKE2/Rancher cluster needs, and the load balancer
+# listener ports (80, 443, 6443, 9345). Set it to reuse a pre-existing group.
+# aws_security_group  = ["sg-"]
 aws_subnet_airgap     = "subnet-0b80e9a553633f5a9" # This subnet should not allow internet connectivity
 aws_subnet_bastion    = "subnet-ee8cac86"
 aws_volume_size       = 500
--- a/tofu/aws/modules/airgap/main.tf
+++ b/tofu/aws/modules/airgap/main.tf
@@ -13,6 +13,82 @@ provider "aws" {
   secret_key = var.aws_secret_key
 }
 
+locals {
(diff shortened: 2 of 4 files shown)


lscalabrini01 (approved):

slickwarren (commented):
lgtm for the most part but how does this tie into the other PRs you have up? They all seem related right?

floatingman:
> lgtm for the most part but how does this tie into the other PRs you have up? They all seem related right?

@slickwarren This is necessary for air-gapped cluster creation in AWS after all of the existing security groups were deleted; the other PR is for publicly accessible clusters.

PR #167, opened 2026-08-07
[Dashboard] - Make a run leave a local dashboard checkout as it found it

izaac:
Running with `--dashboard-dir` against a checkout that has its own root `node_modules` failed in three ways at once, none of which pointed at the cause.

`results.xml` came out empty, because the reporter never loaded:

```
Error loading the reporter: cypress-multi-reporters
Cannot find module '/e2e/node_modules/cypress-multi-reporters'
```

Cypress resolves a reporter against the project root alone. It ignores `NODE_PATH` and does not walk nested `node_modules`, so a real root install makes the reporter unloadable and no JUnit XML is written at all.

The run also modified a tracked file. Installing the packages a branch does not declare writes `cypress/package.json` and leaves the lockfile beside it alone, so the next install rejects a manifest the lockfile does not pin and a second run refuses to start.

Jenkins never saw any of this. A fresh clone has no root `node_modules` and is used once.

## The cause

The run assumed a checkout it owns. It read and wrote paths that belong to the developer: the checkout root for module resolution, a tracked manifest, a `node_modules` link, the Docker build context, and a report directory that is never cleared.

## The fix

- Resolve the reporter and `jrm` through paths the run controls, not the checkout root. `jrm` moves from `npx --no-install`, which resolves only against `./node_modules/.bin`, to `PATH`.
- Snapshot `cypress/package.json` and put it back after installing packages a branch does not declare. Restored on success, on failure and on `SIGTERM`.
- Remove the `node_modules` link the run creates. A developer's own directory or link is never touched.
- Build the test image from a context of its own. `Dockerfile.ci` reads one file, and building from the checkout shipped the whole working tree.
- Clear the JUnit directory at the start of a run. `mocha-junit-reporter` appends, so a reused checkout merged every earlier run's XML.
- Anchor every path the run deletes to a directory verified to hold `cypress/package.json`.
- Install without lifecycle scripts and with the cloud credentials, tokens and `KUBECONFIG` removed from the environment.
- Require `release-2.15` or newer for `--dashboard-dir`, checked before anything is provisioned. Older branches need packages installed into the checkout, which is what a local run must not do. They still run through the clone path.
- Document what a run leaves behind and how to clear it.

The clone path is unchanged.

## Testing

Fifteen local runs against real infrastructure, across `release-2.14`, `release-2.15` and `master`, four root `node_modules` states (absent, a real install, a stale cross branch install, a leftover link) and both execution paths. Each was checked afterwards for a `cypress/package.json` sha that had not moved and for a link left behind.

The state that breaks is a stale cross branch tree, which is what switching branches leaves. A full root install on 2.15 or master happens to satisfy Cypress, because those root manifests carry the reporters, so both variants were run per branch. Twelve component probes covered what a run cannot reach on demand, including the manifest restored on a failed install and on `SIGTERM`, and a non-checkout directory refused.

## Verified on Jenkins

Eight builds, three branches, both `create_initial_clusters` paths. No reporter error, no false positive from the new path guard, and every build recorded its results.

| Build | Branch | Result |
|-------|--------|--------|
| 407 | master | SUCCESS, 53 of 53, 12 JUnit files merged |
| 399 | release-2.15 | SUCCESS, `create_initial_clusters: true` |
| 408 | release-2.14 | 52 of 53, 11 JUnit files merged |

The single failure on `release-2.14` is a spec bug on that branch, not a pipeline one. `get-support.spec.ts` asserts a standard user cannot see the Support link, without `@noPrime`, and those runs used `rancher-prime`. The spec was deleted on `release-2.15` and `master`.

### DIFF
--- a/ansible/testing/dashboard-e2e/README.md
+++ b/ansible/testing/dashboard-e2e/README.md
@@ -415,15 +415,22 @@ This is useful when:
 > are responsible for ensuring that their local code branch is compatible with the targeted
 > `rancher_helm_repo` and `rancher_image_tag`, otherwise tests may fail with unexpected UI errors.
 
-> **Branch requirement:** the checkout must carry its test dependencies in
-> `cypress/package.json` and `cypress/yarn.lock`. That layout exists from
-> `release-2.14` onwards. Older branches such as `release-2.13` keep the test
-> dependencies in the root `package.json`, which lacks the reporters and
-> preprocessor that `cypress.config.jenkins.ts` requires. Because the dependency
-> overlay is intentionally skipped for local checkouts, `--dashboard-dir` cannot
-> be used with them; `run.sh` rejects such a directory up front. Use the clone
-> path instead, which overlays the dependency manifests from
-> `dashboard_overlay_branch`.
+> **Branch requirement:** `--dashboard-dir` requires a `release-2.15` or newer
+> checkout. The checkout has to carry its test dependencies in
+> `cypress/package.json` and `cypress/yarn.lock`, and those manifests have to
+> declare the reporting and tag filtering packages: `cypress-multi-reporters`,
+> `mocha-junit-reporter`, `junit-report-merger`, `find-test-names` and `globby`.
+> `release-2.15` was the first branch to declare all of them. `release-2.14`
+> carries the manifests but not those packages, and `release-2.13` and older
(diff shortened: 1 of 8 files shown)


yonasberhe23 (approved):
LGTM 🚀

PR #169, opened 2026-08-08
[Dashboard] - Install the test dependencies only from the lockfile

izaac:
Test dependencies could enter a run two ways: from `cypress/yarn.lock`, or added at container start for a branch that did not declare them. The second way had no lockfile, so those packages and their dependencies resolved fresh on every run, and the install wrote to a tracked manifest.

`release-2.14` was the only branch that needed it, and rancher/dashboard#18730 has it declaring them itself.

## The change

`yarn install --frozen-lockfile` is now the only way test dependencies enter a run. A checkout that does not declare what the run loads is refused before anything is provisioned, by `run.sh` and by the playbook.

That removes the run time install, the manifest snapshot and restore that went with it, and the `MISSING_RUNTIME_DEPS` plumbing behind them.

## Testing

Two builds on `release-2.14`, the branch this affects, one with `create_initial_clusters` false and one true. Both 52 of 53, the same result and the same 11 merged JUnit files as the run before this change. The one failure is a spec issue on that branch, unrelated.

Neither log mentions installing anything the checkout does not declare, and the new assertion passed on both.

### DIFF
--- a/ansible/testing/dashboard-e2e/README.md
+++ b/ansible/testing/dashboard-e2e/README.md
@@ -415,22 +415,19 @@ This is useful when:
 > are responsible for ensuring that their local code branch is compatible with the targeted
 > `rancher_helm_repo` and `rancher_image_tag`, otherwise tests may fail with unexpected UI errors.
 
-> **Branch requirement:** `--dashboard-dir` requires a `release-2.15` or newer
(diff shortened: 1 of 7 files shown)


yonasberhe23 (approved):
Looks good!

PR #170, opened 2026-08-08
fix(machineconfig): fix AWS credential keys and auto-detect SG IDs

floatingman:
## Summary

Fixes two bugs in the `rancher/cluster` module's `machineconfig` submodule that prevented downstream clusters from provisioning on AWS.

### 1. AWS credential key mismatch

The `machineconfig` submodule read AWS credentials under unprefixed keys (`access_key`/`secret_key`) while the `cloudcredential` submodule and the README sample use the prefixed names (`aws_access_key`/`aws_secret_key`). Since the cluster module passes a single `node_config` object to both, the machine config (node template) silently received `null` for both credential fields.

### 2. Security group IDs treated as names (the primary blocker)

When security groups are passed as **IDs** (`sg-*`), the docker-machine amazonec2 driver treats them as **names** of groups to create. AWS rejects this:

```
Error: InvalidParameterValue: Value (sg-08e8243a8cfbea8a0) for parameter GroupName is invalid.
Group names may not be in the format sg-*.
```

No EC2 instances were ever launched — every node provision job failed immediately, and `tofu apply` timed out waiting for nodes that never came up.

## Changes

- **`tofu/rancher/machineconfig/main.tf`**: Fixed credential keys to `aws_access_key`/`aws_secret_key` (matching `cloudcredential` and the README)
- **`tofu/rancher/machineconfig/main.tf`**: Auto-set `security_group_readonly = true` when `sg-*` IDs are detected, so docker-machine uses existing groups as-is instead of trying to create new ones
- **`tofu/rancher/cluster/README.md`**: Documented the SG name vs ID behavior

## Verification

Reproduced the failure, applied the fix, and ran `tofu apply` end-to-end:
- All 3 EC2 instances launched successfully
- Provision pod logs confirmed: `Found existing security group (qa_default_ipv4) in vpc-bfccf4d7`
- Cluster reached `state=active` (`Apply complete! Resources: 4 added`)

## Test plan

- [x] `tofu validate` passes
- [x] `tofu apply` provisions a working downstream cluster (3 nodes, active state)
- [x] `tofu destroy` cleans up all resources

### DIFF
--- a/tofu/rancher/cluster/README.md
+++ b/tofu/rancher/cluster/README.md
@@ -65,6 +65,8 @@ node_config = {
 
   aws_instance_type = "t3a.2xlarge"
   aws_security_group = ["rancher-nodes"] 
+  # When passing security group IDs (sg-*), set aws_security_group_readonly = true
+  # so docker-machine uses them as-is instead of trying to create new groups.
 
   aws_subnet = "subnet-123"
   aws_availability_zone = "b"
--- a/tofu/rancher/machineconfig/main.tf
+++ b/tofu/rancher/machineconfig/main.tf
@@ -16,7 +16,11 @@ resource "rancher2_machine_config_v2" "rancher2_machine_config_v2" {
       subnet_id                  = try(var.node_config.aws_subnet, null)
       vpc_id                     = try(var.node_config.aws_vpc, null)
       zone                       = try(var.node_config.aws_availability_zone, null)
-      access_key                 = try(var.node_config.access_key, null)
+      access_key                 = try(var.node_config.aws_access_key, null)
+      # When security groups are given as IDs (sg-*), docker-machine must be told
+      # to use them as-is rather than treating them as names of groups to create.
+      # Auto-detect sg-* IDs unless aws_security_group_readonly is set explicitly.
+      security_group_readonly    = try(var.node_config.aws_security_group_readonly, anytrue([for sg in try(var.node_config.aws_security_group, []) : can(regex("^sg-[0-9a-f]+$", tostring(sg)))]) ? true : null)
       block_duration_minutes     = try(var.node_config.aws_block_duration_minutes, null)
       device_name                = try(var.node_config.aws_device_name, null)
       encrypt_ebs_volume         = try(var.node_config.aws_encrypt_ebs_volume, null)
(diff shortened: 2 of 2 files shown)


tofu/rancher/cluster/README.md
```diff
@@ -65,6 +65,8 @@ node_config = {
 
   aws_instance_type = "t3a.2xlarge"
   aws_security_group = ["rancher-nodes"] 
+  # When passing security group IDs (sg-*), set aws_security_group_readonly = true
+  # so docker-machine uses them as-is instead of trying to create new groups.
```
hamistao: Wait I thought we are auto-detecting the aws_security_group fromat precisely so that we don't have to set this when using `sg-*`. My understanding is that the only scenario where we would have to set `aws_security_group_readonly` explicitly is when using existing group NAMES (not ids), am I off here?

tofu/rancher/machineconfig/main.tf
```diff
@@ -16,7 +16,11 @@ resource "rancher2_machine_config_v2" "rancher2_machine_config_v2" {
       subnet_id                  = try(var.node_config.aws_subnet, null)
       vpc_id                     = try(var.node_config.aws_vpc, null)
       zone                       = try(var.node_config.aws_availability_zone, null)
-      access_key                 = try(var.node_config.access_key, null)
+      access_key                 = try(var.node_config.aws_access_key, null)
+      # When security groups are given as IDs (sg-*), docker-machine must be told
+      # to use them as-is rather than treating them as names of groups to create.
+      # Auto-detect sg-* IDs unless aws_security_group_readonly is set explicitly.
+      security_group_readonly    = try(var.node_config.aws_security_group_readonly, anytrue([for sg in try(var.node_config.aws_security_group, []) : can(regex("^sg-[0-9a-f]+$", tostring(sg)))]) ? true : null)
```
hamistao, line 23: Since we are defining our own logic for handling this, isn't it more clear and robust to use `false` instead of null here?

tofu/rancher/cluster/README.md
floatingman: This comment is wrong. I'll fix it. The only time aws_security_group_readonly = true is needed is when reusing existing groups by name.

tofu/rancher/machineconfig/main.tf
floatingman: null is used by every other attribute in the amazonec2_config block, so I don't want to switch conventions here.

